PR #5, opened 2016-06-13
wpan-connman-plugin: Change the parsing of the MeshLocalPrefix value

darconeous:
This commit changes the parsing of dbus message containing the
IPv6MeshLocalPrefix key. The value for this key is currently encoded
as an string (but in earlier build it was encoded as an array).
The new change checks for the type and should work fine with
both encodings.


darconeous:
Merging immediately due to previous internal review.

PR #14, opened 2016-06-27
NCPInstanceBase: Add return status to set_ncp_version_string()

abtink:
This change adds a return value of type `wpantund_status_t` to `NCPInstanceBase::set_ncp_version_string()`. If the new version string does not match the previous value, the function will invoke `ncp_is_misbehaving()` to reinit the NCP and it will return `kWPANTUNDStatus_InvalidArgument`. Otherwise a success status of `kWPANTUNDStatus_Ok` is returned.


abtink:
@darconeous can you please  check this change? I used the `wpantund_status_t` as return type.

darconeous:
👍

PR #16, opened 2016-06-29
Fix xpanid set/get methods on big endian systems

GiedriusM:
Missing code for big endian systems.

As a side remark, it is sometimes unclear where should little/big/host values be used. I think it would be better (more compact and clearer to the reader) to use htole/htobe functions in the future, rather than the custom memcpy based on a define. Also we should check, maybe even add a test, to check spinel compatibility when different endianess architectures are communicating.

### DIFF
--- a/src/wpantund/NetworkInstance.h
+++ b/src/wpantund/NetworkInstance.h
@@ -38,13 +38,15 @@ struct NetworkId {
 	get_xpanid_as_uint64() const
 	{
 		union {
-			uint64_t ret;
+			uint64_t val;
 			uint8_t data[8];
-		};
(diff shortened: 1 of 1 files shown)


darconeous:
👍  LGTM. Will merge after travis gives the thumbs up.

Thanks for helping!

I agree that there definitely needs to be more big-endian testing of this code. I wonder if we could set up some sort of MIPS-BE emulated test environment that could run on travis?

I agree that we should use something like `htobe64()`/`htole64()`. Those functions seem to be Linux-only, but we could provide an implementation if the configure script determines they are missing.

Issue #15, opened 2016-06-29
Backport va_args spinel fixes from openthread


GiedriusM:
I have experienced problem when running spinel pack/unpack. On some platforms data is incorrectly processed, in others segmentation fault occurs.

Reference: https://github.com/openthread/openthread/pull/208

darconeous:
Thanks for reminding me, I'll go ahead and get this ported over.

darconeous:
Should be all fixed on master now.

PR #18, opened 2016-06-29
openthread: Update OpenThread files to latest versions.

darconeous:
This commit updates the files in `third_party/openthread` to the
versions of the files from OpenThread commit `75358c53ec452f5e5`.

Closes issue #15.


darconeous:
@abtink, mind giving this a quick once over, since it includes a few changes to non-openthread code?

abtink:
Sure. Will take a look..

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -183,7 +183,7 @@ SpinelNCPControlInterface::config_gateway(bool defaultRoute, const uint8_t prefi
     const static int kDefaultRouteFlag = 1 << 1;
     const static int kDhcpFlag = 1 << 3; (void)kDhcpFlag;
 
-	spinel_ipv6addr_t addr = {};
+	struct in6_addr addr = {};
 	uint8_t flags = 0;
```
abtink: This is ok/correct. Just curios why the type changes for addr? 
If I am not mistaken `spinel_ipv6adr_t` is `typedef`ed as `struct in6_addr` in spinel.h

darconeous: >If I am not mistaken `spinel_ipv6adr_t` is typedefed as `struct in6_addr` in spinel.h

Not anymore. Had to typedef it to a `struct { uint8_t bytes[16]; }` because `struct in6_addr` isn't available on all OpenThread platforms. See https://github.com/openthread/openthread/issues/121

abtink:
Looks good. 👍

PR #21, opened 2016-07-12
NCPInstanceBase: Detect and report invalid state transitions

darconeous:
This change was deemed necessary because we saw a case where a
device entered the `OFFLINE` state from the `UPGRADING` state, 
which caused all sorts of havoc.

This change adds some additional checks when a state transition
is triggered so that we detect and report invalid state transitions.
In some cases, invalid state transitions are actually outright
prevented, such as transitioning directly from a detached state
(like `FAULT` or `UPGRADING`) to any state other than `UNINITIALIZED`.

All invalid state transitions are clearly indicated to be bugs
in the logs.

### DIFF
--- a/src/wpantund/NCPInstanceBase.cpp
+++ b/src/wpantund/NCPInstanceBase.cpp
@@ -595,22 +595,48 @@ NCPInstanceBase::get_ncp_state()const
 	return mNCPState;
 }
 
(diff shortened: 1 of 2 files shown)


darconeous:
Reviewed internally by @abtink.

PR #23, opened 2016-07-19
Adding support for `energy scan` to core wpantund and wpanctl.

darconeous:
This commit makes the following changes:

- New methods are introduced in `NCPControlInterface` class for
  starting or stopping energy scan.

- It adds new dbus APIs for requesting energy scan and signaling
  the scan results (DBus APIs v1).

- The wpanctl scan command is modified to enable both network
  scan and energy scan, and to be able to parse and output the
  energy scan results.

This commit does not add the implementation of energy scan for
Spinel plugin.


darconeous:
Reviewed internally.

PR #25, opened 2016-07-20
SpinelNCPControlInterface: Fix config_gateway, add/remove external route

abtink:
This commit changes how the `config_gateway`, `add/remove_external_route` are implemented in `SpinelNCPControlInterface` (spinel plugin). It ensures that after an update to any local network data (mesh prefixes or external routes), the new data is registered with the leader. This is done by adding a new Spinel property `THREAD_ALLOW_LOCAL_NET_DATA_CHANGE` with boolean type. This property is set to 'true' before modifying any properties related to local network data. When done, this property is set to 'false' causing the network data to be registered with leader. This commit also adds a new SpinelNCPTask, namely `SpinelNCPTaskChangeNetData` which implements the common logic for changing any net data property.

This commit is related to openthread issue #272.

### DIFF
--- a/src/ncp-spinel/SpinelNCPTaskChangeNetData.cpp
+++ b/src/ncp-spinel/SpinelNCPTaskChangeNetData.cpp
@@ -1,0 +1,115 @@
+/*
+ *
+ * Copyright (c) 2016 Nest Labs, Inc.
+ * All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


abtink:
The related change in openthread is [PR 273](https://github.com/openthread/openthread/pull/273).

src/ncp-spinel/SpinelNCPTaskChangeNetData.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+	require_noerr(ret, on_error);
+
+	mNextCommand = mChangeNetDataCommand;
+
+	EH_SPAWN(&mSubPT, vprocess_send_command(event, args));
+
+	ret = mNextCommandRet;
+
+	require_noerr(ret, on_error);
```
darconeous: If this fails we need to make sure we set `PROP_THREAD_ALLOW_LOCAL_NET_DATA_CHANGE` back to `false`, otherwise we won't be able to make any more changes.

abtink: Indeed. Will change it.

darconeous:
👍

PR #26, opened 2016-07-25
wpan-connman-plugin: Remove scanned networks upon a successful connect

abtink:
This commit changes the wpan-connman plugin so that when NCP state changes to `associated` state, plugin removes all the other networks (services) from the lowpan device.


abtink:
This change was reviewed internally.
@darconeous  can you please help merge this? Thanks.

darconeous:
By the way, we need to fix this since this commit was reverted by commit bb770d9.

abtink:
Perfect! It's good that it's reverted. I have a fix for this but not tested yet.

PR #27, opened 2016-07-28
wpan-connman-plugin: Ensure that network info is retained in case of Connman untimely restart 

abtink:
This commit changes how the disconnect is handled in the wpan-connman-plugin. If a disconnect command/call is received and if this is not user initiated, the current NCP state is checked:
If it is in `ASSOCIATING` or `NCP_STATE_CREDENTIALS_NEEDED` states, instead of issuing a `Leave` command to wpantund, we `Reset` the NCP. This ensures that ongoing association process is stopped, but will not remove/erase any retained NCP network info.

### DIFF
--- a/src/connman-plugin/wpan-connman-plugin.c
+++ b/src/connman-plugin/wpan-connman-plugin.c
@@ -1136,10 +1136,37 @@ lowpan_device_leave(struct connman_device *device)
 		dbus_message_unref(message);
 }
 
+static void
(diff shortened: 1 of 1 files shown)


abtink:
@darconeous can you please help review and merge this. Thanks.

darconeous:
👍 

A few minor style nits, but that file has a lot of inconsistent style anyway—I'm not going to hold this change back due to that.

Will merge once Travis gives the green light.

PR #29, opened 2016-07-29
ncp_spinel: Miscellaneous minor bug fixes

darconeous:
### ncp_spinel: Fetch counters as 32-bit integers, not 8-bit integers.

This commit fixes a copy-paste bug that accidentally fetched
counter values as 8-bit values, when they are actually 32-bit values.
This caused only the least-significant portion of the counter
to be returned.

### ncp_spinel: Tone down the wording on a common error message.

At initialization, we have a list of properties we want to
try and fetch so that we can make sure that the state in
`wpantund` matches that of the NCP. However, not all of these
properties might be implemented, which is OK. In such a case
we log a warning and continue.

The wording of this error message has caused some concern, so
I've reworded the error message and also change the warning
to also lookup the textual representation of both the property
and the error, so that it is clear what exactly happened.

### ncp_spinel: Fault state should be recoverable by reset

It was noticed that when wpantund entered the `uninitialized:fault`
state that the `Reset` command didn't coax it back into the
`uninitialized` state. This change addresses this.

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -121,6 +121,9 @@ SpinelNCPControlInterface::attach(CallbackWithStatus cb)
 void
 SpinelNCPControlInterface::reset(CallbackWithStatus cb)
 {
+	if (mNCPInstance->get_ncp_state() == FAULT) {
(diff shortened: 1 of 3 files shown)


mszczodrak:
LGTM 👍

abtink:
LGTM 👍

PR #30, opened 2016-08-01
ncp_spinel: Log data from PROP_STREAM_DEBUG to SysLog

darconeous:
Instead of writing log messages directly to the serial port, Spinel
has the capability to allow the NCP to write its log messages directly
to the host in-band via the Spinel stream property
`PROP_STREAM_DEBUG`.

This change makes sure that any log messages written to
`PROP_STREAM_DEBUG` get written to the `wpantund` logs.


abtink:
LGTM 👍

darconeous:
I'm pretty sure the build failure was spurious.

PR #31, opened 2016-08-02
ncp_spinel: Fixes for "form" command and permit-join mode.

darconeous:
* 22e1e93 ncp_spinel: Allow spinel driver to enter assisting/permit-join mode.
* 0757865 ncp_spinel: Have "Form" generate PANID, XPANID, and Mesh-local Prefix
* 87a90f7 utils: Added secure random number functions

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
@@ -247,7 +247,41 @@ SpinelNCPInstance::ncp_to_driver_pump()
 		{
 			uint16_t frame_crc = (mInboundFrame[mInboundFrameSize]|(mInboundFrame[mInboundFrameSize+1]<<8));
 			if (mInboundFrameHDLCCRC != frame_crc) {
+
+				int i;
+				static const uint8_t kAsciiCR = 13;
+				static const uint8_t kAsciiBEL = 7;
+
 				syslog(LOG_ERR, "[NCP->]: Frame CRC Mismatch: Calc:0x%04X != Frame:0x%04X, Garbage on line?", mInboundFrameHDLCCRC, frame_crc);
+
+				// This frame might be an ASCII backtrace, so we check to
+				// see if all of the characters are ascii characters, and if
+				// so we dump out this packet directly to syslog.
+
+				mInboundFrameSize += 2;
+
+				for (i = 0; i < mInboundFrameSize; i++) {
+					// Acceptable control codes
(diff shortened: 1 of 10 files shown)


src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
```diff
@@ -247,7 +247,41 @@ SpinelNCPInstance::ncp_to_driver_pump()
...
 				syslog(LOG_ERR, "[NCP->]: Frame CRC Mismatch: Calc:0x%04X != Frame:0x%04X, Garbage on line?", mInboundFrameHDLCCRC, frame_crc);
+
+				// This frame might be an ASCII backtrace, so we check to
+				// see if all of the characters are ascii characters, and if
+				// so we dump out this packet directly to syslog.
+
+				mInboundFrameSize += 2;
+
+				for (i = 0; i < mInboundFrameSize; i++) {
+					// Acceptable control codes
```
arjunasiva: subjective style nit: I tend to like to have a temporary for accesses into an array. I.e:
`uint8_t currentByte = mInboundFrame[i];`

It's all the same to the compiler but I've always been found it more readable

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -43,6 +43,33 @@ using namespace wpantund;
 
 WPANTUND_DEFINE_NCPINSTANCE_PLUGIN(spinel, SpinelNCPInstance);
 
+void
+SpinelNCPInstance::handle_ncp_log(const uint8_t* data_ptr, int data_len)
+{
+    static char linebuffer[NCP_DEBUG_LINE_LENGTH_MAX + 1];
+    static int linepos = 0;
+    while (data_len--) {
```
arjunasiva: nit: style. You've got a `for(...)` loop above but a while loop for iterating through every byte in a buffer here.

arjunasiva:
👍

src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
darconeous: I agree, not sure why I didn't do that in this case.

src/ncp-spinel/SpinelNCPInstance.cpp
darconeous: It's because I can get away with mutating the arguments here, whereas I didn't want to mutate the arguments in the for loop above.

PR #33, opened 2016-08-08
Adding NetworkRetain class and wpanretain script

abtink:
This commit adds a new class `NetworkRetain` to wpantund. This class provides hooks to enable retaining (saving/recalling) network info. This is realized by adding a new configuration property to wpantund `Config:Daemon:NetworkRetainCommand`. This property provides the command to perform the network retain operations. The `NetworkRetain` class executes the retain command from a forked child process to request  a save operation (when NCP state transition to "associated" ) or an erase (upon a `leave`) or network info recall and restore (upon NCP reset or a wpantund restart). The `NetworkRetain` uses the same model as in `FirmwareUpgrade` module (i.e., the forked process
is started before wpatund possibly dropping privileges).

This commit also adds a script `wpanretain.sh` which can be used as the network retain command. This script implements the save, recall and erase operations. It uses `wpanctl get property` commands and
parses the responses to collect network parameters to save in a config file.

### DIFF
--- a/src/scripts/wpanretain.sh
+++ b/src/scripts/wpanretain.sh
@@ -1,0 +1,346 @@
+#
+# Copyright (c) 2016 Nest Labs, Inc.
+# All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


darconeous:
This change looks good to me for now, but I think as a follow up change we should consider having separate commands for `save`, `recall`, and `erase`. The script you have would remain the same, you would just specify the argument as a part of the command.

I think that would be better in the long run since it places no restrictions on how the command is supposed to behave. It doesn't have to worry about what `wpantund` might pass to it because it just executes the command as given.

I'm open to being convinced otherwise, and I don't feel strongly enough about it to hold back this commit. So... 👍 .

If you are curious, I feel the same way about the version check, it's just that the only alternative in that case feels even more hokey than passing an argument: Setting an environment variable. Bleh.

abtink:
Thinking about it I fully agree. Separating it into 3 commands makes perfect sense (and make it more flexible).  Will modify this later when I get the chance.

PR #35, opened 2016-08-11
NCPInstanceBase: Allow FAULT and UPGRADING states to be reported, even when initializing.

darconeous:
If this isn't allowed, we might never be able to query the NCP for status if it enters the FAULT state.

This pull request also include the following commit:

* ncp: Stop printing out `get_property` debug prints during initialization.

    During initialization, logs were being polluted with lines like:

        wpantund[12345] <info> get_property: key: "NCP:State"
        wpantund[12345] <info> get_property: key: "NCP:State"
        wpantund[12345] <info> get_property: key: "NCP:State"

    These lines are useless, and come from the the main wpantund run-loop
    attempting to determine if the NCP instance is ready to be announced
    via D-Bus. However, this log line is still generally useful, this
    commit adjusts the behavior to only print this message when we are not
    initializing.

### DIFF
--- a/src/ncp-dummy/DummyNCPControlInterface.cpp
+++ b/src/ncp-dummy/DummyNCPControlInterface.cpp
@@ -217,7 +217,9 @@ DummyNCPControlInterface::get_property(
     const std::string& in_key, CallbackWithStatusArg1 cb
     )
 {
-	syslog(LOG_INFO, "get_property: key: \"%s\"", in_key.c_str());
+	if (!mNCPInstance->is_initializing_ncp()) {
+		syslog(LOG_INFO, "get_property: key: \"%s\"", in_key.c_str());
+	}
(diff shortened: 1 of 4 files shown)


src/ncp-dummy/DummyNCPControlInterface.cpp
```diff
@@ -217,7 +217,9 @@ DummyNCPControlInterface::get_property(
     const std::string& in_key, CallbackWithStatusArg1 cb
     )
 {
-	syslog(LOG_INFO, "get_property: key: \"%s\"", in_key.c_str());
+	if (!mNCPInstance->is_initializing_ncp()) {
+		syslog(LOG_INFO, "get_property: key: \"%s\"", in_key.c_str());
+	}
```
abtink: Nice! I :+1:

abtink:
LGTM :+1:

PR #39, opened 2016-08-12
ncp-spinel: Updates to Spinel API from OpenThread.

darconeous:
This change is required to have `wpantund` successfully control
OpenThread NCPs after [OpenThread pull request #372][1] is merged
in.

Versions of `wpantund` that include this change *should* be able to
continue to control older NCPs, but no guarantees are made.

[1]: https://github.com/openthread/openthread/pull/372


abtink:
LGTM :+1:

PR #41, opened 2016-08-16
Refine and clarify how MAC addresses are handled.

darconeous:
This change updates wpantund to introduce and more clearly define the
following properties that handle MAC addresses:

*   `NCP:MACAddress`: Represents the current MAC address as set by the
    hardware for inbound and outbound traffic. This property may be
    written to, if such a capability is supported by the NCP. This
    value reverts to the NCP default upon reset.
*   `NCP:HardwareAddress`: This is the MAC address that is permanently
    assigned to the device by the manufacturer. It is read-only. Some
    non-Thread NCP implementations may use this as the default value
    for `NCP:MACAddress`.
*   `NCP:ExtendedAddress`: This is the MAC address that is generated
    randomly by the NCP and stored in non-volatile storage for future
    use. Thread NCPs will use this value as the default value for
    `NCP:MACAddress`. This property may be written to, if such a
    capability is supported by the NCP.

Since we were previously using the term "hardware address" to refer to
what was actually a MAC address, various function names have been
accordingly renamed.


abtink:
LGTM :+1:

PR #44, opened 2016-08-18
Spinel.h: Sync spinel header file with openthread

abtink:
This syncs `spinel.h` with openthread at commit adba70a3906.


darconeous:
👍

arjunasiva:
LGTM

PR #46, opened 2016-08-23
ncp-spinel: Add addresses for on-mesh SLAAC prefixes.

darconeous:
This commit makes sure that wpantund adds the appropriate addresses to the interface as on-mesh prefixes become available.


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -1033,6 +1060,46 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
...
+		start_new_task(boost::shared_ptr<SpinelNCPTask>(
+			new SpinelNCPTaskChangeNetData(
+				this,
+				NilReturn(),
+				commands
+			)
+		));
+	}
+}
+
```
abtink: What if an on-mesh prefix is removed and we have previously added an address based on this prefix, should we consider to remove such addresses?

darconeous: I'm still writing that part. It will likely be included with the generalized address list changes. This commit just makes sure that addresses are at least added properly, so that additional testing can get unblocked.

abtink:
LGTM :+1:

PR #49, opened 2016-08-30
wpantund: Refactor how the address list is handled.

darconeous:
`wpantund` has traditionally treated the link local address and the
mesh-local/realm-local addresses as special and handled them independently
from other addresses. I now believe this was a mistake.

This change unifies how all addresses are handled into one single list
that addresses get added to and removed from. This approach simplifies
the internal API and also makes general address housekeeping much easier.

This is a fairly invasive change, so extra care should be taken
in its review.


darconeous:
This doesn't include the prefix management stuff (mentioned [here](https://github.com/openthread/wpantund/pull/46#discussion_r75786777), will need to add that as a separate pull request. (Still not yet written)

abtink:
Went through all changes. LGTM 👍

PR #50, opened 2016-08-31
ncp-spinel: Add support for in-band assisted joining

darconeous:
This commit adds support for in-band assisted joining. This capability
allows a node (which does not yet have network credentials) to communicate
in a very limited fashion (a single application-defined port using IPv6
link-local addresses) with an "assisting device" (Joinable bit set).
Using an undefined application protocol, the joining device can learn the
network credentials from the assisting device and then proceed to join the
network normally.

This change relies on the changes from [OpenThread Pull Request #497][1].

[1]: https://github.com/openthread/openthread/pull/497

### DIFF
--- a/src/ncp-spinel/SpinelNCPTaskJoin.cpp
+++ b/src/ncp-spinel/SpinelNCPTaskJoin.cpp
@@ -44,7 +44,9 @@ void
 nl::wpantund::SpinelNCPTaskJoin::finish(int status, const boost::any& value)
 {
 	SpinelNCPTask::finish(status, value);
-	if (!ncp_state_is_associated(mInstance->get_ncp_state())) {
+	if ( (status != kWPANTUNDStatus_InProgress)
+	  && !ncp_state_is_associated(mInstance->get_ncp_state())
+	) {
 		mInstance->change_ncp_state(mLastState);
 	}
 }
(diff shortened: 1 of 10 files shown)


darconeous:
I'm not entirely satisfied with this commit yet, but I wanted to bring it up to start a discussion.

darconeous:
I think these changes are good for now, so I think they are ready for review.

darconeous:
Waiting on a review.

src/ncp-spinel/SpinelNCPTaskJoin.cpp
```diff
@@ -255,11 +270,23 @@ nl::wpantund::SpinelNCPTaskJoin::vprocess_event(int event, va_list args)
...
-	ret = kWPANTUNDStatus_Ok;
+	ret = last_status
+		? WPANTUND_NCPERROR_TO_STATUS(last_status)
+		: kWPANTUNDStatus_Ok;
+
+	if ( (last_status == SPINEL_STATUS_JOIN_SECURITY)
+	  || (last_status == SPINEL_STATUS_JOIN_FAILURE)
+	) {
+		mInstance->change_ncp_state(CREDENTIALS_NEEDED);
+		ret = kWPANTUNDStatus_InProgress;
```
abtink: Why `InProgress` status here? e.g., if key is not correct we will get the `JOIN_SECURITY` status response, and the join operation was a failure, right?

abtink:
Overall looks good. 👍

src/ncp-spinel/SpinelNCPTaskJoin.cpp
darconeous: The wpantund join command is defined to return `InProgress` if we have switched to the `credentials-needed` state. The join failed, but by supplying the appropriate credentials it will try again.

PR #51, opened 2016-09-02
NcpTypes: Enable conversion of new "node type" strings.

abtink:
This commit enhances the `string_to_node_type()` to allow some more commonly used strings to be used as node type (e.g., "sed" or "sleepy" for sleepy-end-device).


darconeous:
👍

PR #52, opened 2016-09-09
wpanctl: Print node-type in `join`/`form` commands

abtink:
This commits adds to output of `wpanctl` `join` and `form` commands to inlcude the `node-type`. It also modifies the `arg.h` to include the arg type in the printout and adds info in help for valid node types.


darconeous:
👍

PR #54, opened 2016-09-10
Adding support for getting child table

abtink:
This commit adds support to spinel plugin for getting the child table property `SPINEL_PROP_THREAD_CHILD_TABLE`. It also adds a new read-only wpantund property `Thread:ChildTable` to retrieve the child table along with info about each child.

This is the counterpart of the openthread PR [#515](https://github.com/openthread/openthread/pull/515).


arjunasiva:
lgtm 👍

PR #55, opened 2016-09-14
Add support for Energy Scan operation in Spinel plugin

abtink:
This commit adds support to spinel plugin for energy scan operation, specifically it modifies the `SpinelNCPTaskScan` to handle both netscan (active scan) and energy scan operations.

The counterpart of this change in openthread is in [PR580](https://github.com/openthread/openthread/pull/580).


darconeous:
👍

PR #59, opened 2016-09-22
Add ON_MESH_NETs and LOCAL_ROUTs with stable flag set to true

abtink:
This commit changes the `SpinelNCPControlInterface::config_gateway` implementation so that when an `ON_MESH_PREFIX` is added the stable flag is set to `true` by default.


darconeous:
The OS X build failure was spurious (Tracked by #61). Will go ahead and merge.

PR #60, opened 2016-09-22
wpantund: Support for raw packet capture (Packet sniffing)

darconeous:
This pull request adds support for the capture and logging of raw packets from the radio, assuming the underlying NCP supports it.

Note that this feature captures raw packets from the radio, which will currently always be 802.15.4 frames (but that may not always be the case in the future). If you just want to sniff IP traffic, you can simply use Wireshark or tcpdump directly.

Packets are captured in the PCAP format, which is the same format used by Wireshark and tcpdump. You can start a packet capture using the `pcap` command in `wpanctl`:

    Syntax:
       pcap [args] <capture-file>
    Options:
       -h/--help                 Print Help
       -t/--timeout              Set timeout period [ms]
       -f                        Force use of stdout, even if it is a tty

If you to not specify a file, `stdout` is used. If `stdout` is a TTY, it will refuse unless you also specify `-f`.

For real-time monitoring in Wireshark, you can capture to a named pipe (created using `mkfifo`) and have Wireshark capture from that.

Note that support for promiscuous sniffing is not yet supported. That feature will be added in a subsequent pull request.

Raw packet capture is being added to OpenThread NCPs via [PR #644][1].

[1]: https://github.com/openthread/openthread/pull/644

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -475,6 +475,29 @@ SpinelNCPControlInterface::get_ncp_instance()
 	return (*mNCPInstance);
 }
 
+void
+SpinelNCPControlInterface::pcap_to_fd(int fd, CallbackWithStatus cb)
+{
+	int ret = mNCPInstance->mPcapManager.insert_fd(fd);
+
+	if (ret < 0) {
+		syslog(LOG_ERR, "pcap_to_fd: Failed: \"%s\" (%d)", strerror(errno), errno);
+
+		cb(kWPANTUNDStatus_Failure);
+
+	} else {
+		cb(kWPANTUNDStatus_Ok);
+	}
+}
+
+void
+SpinelNCPControlInterface::pcap_terminate(CallbackWithStatus cb)
+{
+	mNCPInstance->mPcapManager.close_fd_set(mNCPInstance->mPcapManager.get_fd_set());
+	cb(kWPANTUNDStatus_Ok);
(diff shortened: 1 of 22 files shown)


src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -113,6 +113,13 @@ SpinelNCPControlInterface::attach(CallbackWithStatus cb)
 		new SpinelNCPTaskSendCommand(
 			mNCPInstance,
 			NilReturn(),
+			SpinelPackData(SPINEL_FRAME_PACK_CMD_PROP_VALUE_GET, SPINEL_PROP_NET_ROLE)
+		)
```
abtink: @darconeous, is this related to pcap change or just a separate change?

src/wpantund/Pcap.cpp
```diff
@@ -0,0 +1,375 @@
...
+			*max_fd = fd;
+		}
+	}
+
+	return 0;
+}
+
+void
+PcapManager::process(void)
+{
```
abtink: In this PR, I cannot find where this is called. I guess maybe from `NCPInstanceBase::Process()`?

src/ncp-spinel/SpinelNCPControlInterface.cpp
darconeous: This was a change that accidentally snuck in, I'll pull it.

darconeous:
Build failure is unrelated. (Tracked by #61)

PR #61, opened 2016-09-22
Fix for sudden onset of travis OS X failures

darconeous:
This issue is tracking the fix for the recent onset of travis failures when building for OS X.


darconeous:
This is largely a bandaid until we get the OS X build working again.

PR #64, opened 2016-09-23
Ensure stable flag is true for on-mesh-prefix added from `address_was_added()`

abtink:
This commit changes the `SpinelNCPInstance::address_was_added()` so that when an `ON_MESH_PREFIX` added the stable flag is set to `true`.


darconeous:
Build failure is unrelated. (Tracked by #61)

arjunasiva:
nice catch! LGTM

PR #69, opened 2016-09-27
ncp-spinel: Consolidate and refine behavior of SendCommand task

darconeous:
This change takes the functionality that is currently in the
ChangeNetData task and merges it into the SendCommand task. It
also introduces a new factory design pattern which makes instantiating
the task more straightforward and easy to use.


src/ncp-spinel/SpinelNCPTaskSendCommand.cpp
```diff
@@ -210,42 +263,82 @@ nl::wpantund::SpinelNCPTaskSendCommand::vprocess_event(int event, va_list args)
...
+			mLockProperty,
+			true
+		);
+
+		EH_SPAWN(&mSubPT, vprocess_send_command(event, args));
+
+		mRetVal = mNextCommandRet;
+
+		// In case of BUSY error status (meaning the `ALLOW_LOCAL_NET_DATA_CHANGE`
+		// was already true), allow the operation to proceed.
```
abtink: Suggestion: "meaning `mLockProperty` was already true...." instead of "meaning the `ALLOW_LOCAL_NET_DATA_CHANGE`"

src/ncp-spinel/SpinelNCPTaskSendCommand.cpp
```diff
@@ -31,17 +31,71 @@
...
+{
+	mCb = cb;
+	return *this;
+}
+
+SpinelNCPTaskSendCommand::Factory&
+SpinelNCPTaskSendCommand::Factory::set_callback(const CallbackWithStatus &cb)
+{
+	mCb = boost::bind(cb, _1);
+	return *this;
```
abtink: Nice!

PR #73, opened 2016-09-29
Adding new wpantund properties `Thread:RLOC16` and (write-only) `Thread:PreferredRouterID`

abtink:
Adding new wpantund properties `Thread:RLOC16` and (write-only) `Thread:PreferredRouterID`


abtink:
The new commit adds a new property `Thread:RouterId` in addition to other two.

abtink:
@darconeous , can you please help merge this?
It's also already tested/verified.

PR #78, opened 2016-10-07
autoandr: Add automatic generation of Android.mk files

darconeous:
This commit introduces a new shell script which allows us to automatically generate `Android.mk` files (needed to build for Android and Brillo platforms) from the autotools files.

The tool is currently experimental.

To use it, either set `NDK_ROOT` to point to your Android NDK (or make sure that the `ndk-build` tool is in your executable path) and then add the `--all` option to the bootstrap script:

	./bootstrap.sh --all


darconeous:
I'm inclined to go ahead and pull the trigger on this so that people start playing with it, since the capability is experimental and is largely unrelated to `wpantund` functionality.

PR #79, opened 2016-10-13
ncp-spinel: Updates for pcap support

darconeous:
OpenThread Issue [#809][1] addresses some issues with the `PROP_STREAM_RAW` interface. This is the `wpantund`-side of those changes.

[1]: https://github.com/openthread/openthread/pull/809


darconeous:
Going ahead and merging this since I have verified that the changes behave properly.

PR #82, opened 2016-10-18
Measure prefix length in bits, not bytes

darconeous:
As a carry-over from days long gone, several code paths in `wpantund` measure the prefix length in bytes, not bits. This change addresses this issue.

Also includes improvements for a common error condition (interface not found) in `wpanctl`.

### DIFF
--- a/src/wpanctl/tool-cmd-config-gateway.c
+++ b/src/wpanctl/tool-cmd-config-gateway.c
@@ -60,6 +60,10 @@ int tool_cmd_config_gateway(int argc, char* argv[])
 	uint32_t preferredLifetime = 0xFFFFFFFF;
 	uint32_t validLifetime = 0xFFFFFFFF;
 	const char* prefix = NULL;
+	uint8_t prefix_length = 64;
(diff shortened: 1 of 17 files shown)


src/wpanctl/tool-cmd-config-gateway.c
```diff
@@ -201,24 +204,17 @@ int tool_cmd_config_gateway(int argc, char* argv[])
...
-			    DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, "", 0,
-			    DBUS_TYPE_INVALID
-			);
 		}
 
+		addr = prefix_bytes;
+
 		dbus_message_append_args(
 		    message,
+			DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &addr, 16,
```
abtink: nit: I think this is tab vs space, `DBUS_TYPE_ARRAY` does not look aligned with the rest.

abtink:
I was thinking at some point we also can consider changing the add/remove-route DBUS APIS to give the route prefix in bits instead of bytes...

PR #94, opened 2016-11-08
Add child table and last status logs

abtink:
This commit changes the `TaskGetChildTable` to expose some new methods (to parse/get the child table from the corresponding spinel message). It also adds logs
- when `LAST_STATUS` is received (the status)
- when `CHILD_TABLE` property changes is received (prints the entire
  child table in the syslogs).

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -662,6 +662,7 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
 	if (key == SPINEL_PROP_LAST_STATUS) {
 		spinel_status_t status = SPINEL_STATUS_OK;
 		spinel_datatype_unpack(value_data_ptr, value_data_len, "i", &status);
+		syslog(LOG_INFO,"[-NCP-]: Last status (%s, %d)", spinel_status_to_cstr(status), status);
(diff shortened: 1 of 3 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -662,6 +662,7 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
 	if (key == SPINEL_PROP_LAST_STATUS) {
 		spinel_status_t status = SPINEL_STATUS_OK;
 		spinel_datatype_unpack(value_data_ptr, value_data_len, "i", &status);
+		syslog(LOG_INFO,"[-NCP-]: Last status (%s, %d)", spinel_status_to_cstr(status), status);
```
darconeous: Perhaps include the TID?

abtink: Good idea, but here, in `handle_ncp_spinel_value_is()`, we don't have the full spinel message and its header (we get the key and value as a data blob), so adding `tid` here won't be straight forward.

But we get the `tid` in the log from `handle_ncp_spinel_callback()`.
With this change the logs look like this (first line show we got a LAST_STATUS and the tid, second line shows the status string and code): 

> wpantund[141019]: [NCP->] CMD_PROP_VALUE_IS(PROP_LAST_STATUS) tid:10
>  wpantund[141019]: [-NCP-]: Last status (STATUS_OK, 0) 
>

PR #43, opened 2016-08-17
Add diagnostics(mfg) interface in wpantund

chshu:
- Spinel: add 'SPINEL_PROP_NEST_STREAM_MFG' property

  - D-Bus: add 'WPANTUND_IF_CMD_MFG' command

  - Move original 'NCPMfgInterface' to 'NCPMfgInterface_v0', add new 'NCPMfgInterface_v1'

PR #98, opened 2016-11-09
wpanctl: Added `mfg` command.

darconeous:
This is needed to execute the manufacturing commands on OpenThread.


darconeous:
Somehow this change got left out from #43.

PR #109, opened 2016-12-01
Add support for new property "JamDetection:Debug:HistoryBitmap"

abtink:
This commit adds a new wpan property to get the jam detection history bitmap. This value provides information about current state of jamming detection module for monitoring/debugging purpose. It returns a 64-bit value where each bit corresponds to one second interval starting with bit 0 for the most recent interval and bit 63 for the oldest intervals (63 sec earlier). The bit is set to 1 if the jamming detection module observed/detected high signal level during the corresponding one second interval.

(This commit also fixes some of the indention in `SpinelNCPInstance.cpp` to use tabs everywhere). 

The counter-part of this change in OpenThread: https://github.com/openthread/openthread/pull/1033

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -49,28 +49,28 @@ WPANTUND_DEFINE_NCPINSTANCE_PLUGIN(spinel, SpinelNCPInstance);
 void
 SpinelNCPInstance::handle_ncp_log(const uint8_t* data_ptr, int data_len)
 {
-    static char linebuffer[NCP_DEBUG_LINE_LENGTH_MAX + 1];
-    static int linepos = 0;
-    while (data_len--) {
-        char nextchar = *data_ptr++;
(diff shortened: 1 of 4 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -271,6 +271,7 @@ SpinelNCPInstance::get_supported_property_keys()const
 		properties.insert(kWPANTUNDProperty_JamDetectionRssiThreshold);
 		properties.insert(kWPANTUNDProperty_JamDetectionWindow);
 		properties.insert(kWPANTUNDProperty_JamDetectionBusyPeriod);
+        properties.insert(kWPANTUNDProperty_JamDetectionDebugHistoryBitmap);
```
darconeous: Whitespace.

abtink: Ah, yes. Thanks. I fixed it in new commit push .

PR #117, opened 2017-01-09
Network Topology Tables as ValueMaps 

aeliot:
Added support for getting the topology tables (neighbor table and child table) in ValueMap format rather then string format.


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -490,6 +499,16 @@ SpinelNCPInstance::get_property(
 			)
 		));
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_ThreadNeighborTableAsValMap)) {
```
abtink: nit: spacing alignment (probably tabs vs spaces)

src/ncp-spinel/SpinelNCPTaskGetNetworkTopology.cpp
```diff
@@ -333,6 +333,36 @@ SpinelNCPTaskGetNetworkTopology::TableEntry::get_as_string(void) const
...
+    addr |= (uint64_t) mExtAddress[6] << 8;
+    addr |= (uint64_t) mExtAddress[5] << 16;
+    addr |= (uint64_t) mExtAddress[4] << 24;
+    addr |= (uint64_t) mExtAddress[3] << 32;
+    addr |= (uint64_t) mExtAddress[2] << 40;
+    addr |= (uint64_t) mExtAddress[1] << 48;
+    addr |= (uint64_t) mExtAddress[0] << 56;
+
+    entryMap.insert( std::pair<std::string, boost::any>( "extAddress",     addr               ));
+    entryMap.insert( std::pair<std::string, boost::any>( "rolc16",         mRloc16            ));
```
abtink: I suggest that we define all these strings in a header file somewhere (maybe in wpan-properties), something like this: 
```
#define kWPANTUNDValueMapKey_ChildNeighborEntry_ExtAddress        "ExtAddress"
#define kWPANTUNDValueMapKey_ChildNeighborEntry_Rloc16            "RLOC16" 
#define kWPANTUNDValueMapKey_ChildNeighborEntry_LinkQualityIn     "LinkQualityIn"
...
```
This way there is one place they are defined and the same `#define`s can later be used in sapphire code.

aeliot: Do you think the get_as_string function above should also use the same names?  The keys are currently the same as the names used when outputting as a string

abtink: Maybe... would be good to change it but not critical...

PR #124, opened 2017-01-25
Check for success when adding address/on-mesh-prefix

abtink:
This commit adds a callback in `address_was_added()` when issuing
the set the spinel commands to check for success of the whole
operation.


abtink:
Closing and replacing this with #124

PR #129, opened 2017-02-01
build: Include a stub copy of boost for systems that don't have it.

darconeous:
We've run into trouble with the dependency on [Boost](http://boost.org) recently, and since the parts we are using are entirely implemented as template headers we figured the easiest path forward was to simply include those headers with wpantund.

On systems which already have Boost installed, that version will be used. The bundled version will only be used when no usable copy of Boost can be found.


beriberikix:
Did you mean @jwhui?

darconeous:
Oops, extra click.

darconeous:
Trying to figure out the travis build failure.

PR #131, opened 2017-02-09
`TaskLeave`: Issue a "reset" at the end of "leave"

abtink:
This ensures that all the volatile memory on the NCP is wiped after
a network "leave" command.


arjunasiva:
LGTM 👍

PR #134, opened 2017-02-27
Add initial wpanctl commissioner commands support

pszkotak:
This pull request adds support for wpanctl commands to use border router as a native commissioner and joining a device to the the Thread network.


src/ipc-dbus/DBusIPCAPI_v1.cpp
```diff
@@ -1192,6 +1196,76 @@ DBusIPCAPI_v1::interface_route_remove_handler(
...
+	const char* psk = NULL;
+	int psk_len = 0;
+	bool did_succeed = false;
+
+	did_succeed = dbus_message_get_args(
+		message, NULL,
+		DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &ext_addr, &ext_addr_len,
+		DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &psk, &psk_len,
+		DBUS_TYPE_INVALID
+);
```
abtink: nit: spacing here

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -309,6 +310,62 @@ SpinelNCPControlInterface::add_external_route(
...
+			SPINEL_DATATYPE_EUI64_S
+			SPINEL_DATATYPE_DATA_S
+			SPINEL_DATATYPE_UINT8_S
+		),
+		SPINEL_PROP_THREAD_JOINERS,
+		addr,
+		psk,
+		psk_len
+	))
+	.set_lock_property(SPINEL_PROP_THREAD_ALLOW_LOCAL_NET_DATA_CHANGE)
```
abtink: nit: spacing here ... it'd better to have all the factory method calls aligned...

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -309,6 +310,62 @@ SpinelNCPControlInterface::add_external_route(
...
+	.finish()
+	);
+
+bail:
+	return;
+}
+
+void
+SpinelNCPControlInterface::commissioner(
```
abtink: Suggestion: Since this is just to enable/disable commissioner operation, I think it may be easier if instead of adding it as a `NCPControlInterface`  new method, we define it as a wpan property (which is then tied to the spinel property... `SPINEL_PROP_THREAD_COMMISSIONER_ENABLED`)...Thoughts?

pszkotak: Thank you, it looks way better. Changed to the wpan property. I think that support for commissioner --start / commissioner --stop commands should be present. Using wpanctl setprop / getprop and then commissioner --joiner-add does not seem to be very intuitive. Therefore --start / --stop which set spinel property are still there. Do you think such redundancy is acceptable in this case?

pszkotak:
Please let me know if there are any more improvements needed before merging this PR.

GiedriusM:
I'm really missing this functionality from upstream wpantund, so I've merged this PR locally.
Here's few things I've noticed:
* --state always return an error 10 (0x0A)
* it's hard to switch between "commissioner --state" and "status", i.e. frequent mistakes of "commissioner --status" and "state"
* some commands (see below) return no information and it's unclear what has been done
* some commands (see below) return errors
* stopping commissioner does not clear joiner list
* stopping commissioner does not remove joinable flag in scan entries

```
wpanctl:wpan0> commissioner
wpanctl:wpan0> commissioner abcdefgh ffffffffffffffff 600
wpanctl:wpan0> commissioner abcdefgh
wpanctl:wpan0> commissioner abcdefgh ffffffffffffffff
```

```
wpanctl:wpan0> commissioner -t 10 -a abcdefgh
commissioner: warning: No address value specified, any joiner knowing PSKd can join.
commissioner: warning: PSK should consist only uppercase letters and digits
commissioner: error: Did not receive a reply. Possible causes include: the remote application d
id not send a reply, the message bus security policy blocked the reply, the reply timeout expir
ed, or the network connection was broken.
Error 11 (0x0B)
```

GiedriusM:
One more issue - "wpanctl commissioner -a PASSW0RD" prints "Joiner added." to stderr instead of stdout.

darconeous:
@abtink: What are your thoughts on this? Do you approve?

darconeous:
### Note to self: This commit has a lot of unnecessary history, so don't "Merge", do "Squash and Merge" instead.

darconeous:
Ignore googlebot.

src/wpanctl/tool-cmd-commissioner.c
```diff
@@ -0,0 +1,367 @@
...
+                    print_error_diagnosis(ret);
+                }
+            }
+        }
+    }
+
+
+bail:
+    if (connection)
+        dbus_connection_unref(connection);
```
darconeous: Due to the potent example of the ["goto fail"](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/) bug, we now require curly braces for `if` (and all similar constructs) in all new code.

src/wpanctl/tool-cmd-commissioner.c
```diff
@@ -0,0 +1,367 @@
...
+                     "%s/%s",
+                     WPANTUND_DBUS_PATH,
+                     gInterfaceName);
+
+            message = dbus_message_new_method_call(
+                interface_dbus_name,
+                path,
+                WPANTUND_DBUS_APIv1_INTERFACE,
+                WPANTUND_IF_CMD_PROP_SET
+                );
```
abtink: nit: ");" align with `message`.  Few other instances below.

darconeous:
I'm not sure whether I should wait for @pszkotak to address those nits, or if I should go ahead and merge. I'm leaning toward waiting for now. If someone feels differently, let me know.

pszkotak:
@GiedriusM,
Thank you for the detailed feedback.
We are still missing some functionalities, like 'state' returning ERRORCODE_NOT_IMPLEMENTED.
It is a good idea to implement everything that is defined before the merge.
Will check all remaining points as well.

@darconeous, sorry for the delay.
I will address all nits asap.

darconeous:
Note that you should do a pull to your local copy before making changes, since I've made a few.

pszkotak:
@GiedriusM,
1. --state, which is now renamed to --status to match 'wpanctl status' cmd, has beed added, it returns enabled/disabled string
2. If there is no confirmation nothing has been done.
3. These two are more openthread related.
I cant find in spec if the joiner list should be cleared when the commissioner is stopped.
Could you point me the right direction? ;)
We will look into the j flag behavior, it may be a bug. Anyway, if there is anything to be fixed it is not in wpantund/wpanctl.
4. -t 10 is too short, this is why it is failing. Please note that the default timeout value equals 60000ms.
```
wpanctl commissioner -t 100 -a ABCDEFGH
commissioner: warning: No address value specified, any joiner knowing PSKd can join.
Joiner added.
```
5. Writing command result to the stderr seems to be a common practice among all tool-cmd-*.c files, just followed a convention here:
e.g. tool-cmd-add-route.c
`fprintf(stderr, "Adding route prefix \"%s\" ...",`
If there is a need to change it all commands should be adjusted at the same time imo.

@darconeous, there is still --remove command missing, but as this requires also changes in the opentherad. I think it can be added with one of future pull requests as complementary to this one has already been merged. If all looks good here I will stop messing with this PR ;).

darconeous:
I hope to review and approve these changes today.

darconeous:
I will fix the conflicts

PR #135, opened 2017-02-28
Add support for new property to test assert functionality on NCP

abtink:
- Add a new wpan property "OpenThread:Debug:TestAssert" which maps
  to `SPINEL_PROP_DEBUG_TEST_ASSERT`.

- Reading/Getting this property will cause an assert on NCP.
  This is intended for testing the assert functionality of
  NCP. Assert should ideally cause the NCP to reset, but if this is
  not supported a `false` boolean is returned in response.


abtink:
Corresponding PR in OT: https://github.com/openthread/openthread/pull/1389

darconeous:
If we can get #141 in first, then we can remove the spinel changes.

abtink:
Rebased on master.

darconeous:
Spinel should still define the constants though. The syslog enumeration won't ever change, it is enshrined in RFCs, but we shouldn't require people to include syslog.h.

darconeous:
Wow, that last comment I made was totally off the wall and applied to a different pull request (even then it wasn't accurate). I've deleted it. If you see it in your email, disregard it, it will only confuse you. :P

PR #141, opened 2017-03-13
Sync OpenThread files to latest versions and update plugin to match

darconeous:
This pull request includes two changes, one is the sync of files that have changed in OpenThread and the other updates various parts of the Spinel plug-in to reflect those changes.

The later commit is larger than expected due to the realization that `spinel-extra.c` was quite crufty. The code in question wasn't being used in the ways in which it was broken, which is why we haven't seen any problems so far. Nonetheless, it seemed prudent to fix the code.

This change also includes the latest updates ([#1441](https://github.com/openthread/openthread/pull/1441)) to `spi-hdlc-adapter`, previous versions of which had a nasty frame corruption bug.


darconeous:
Adding some additional reviewers since @abtink  is out this week.

darconeous:
I've done some testing and these changes seem to work fine.

PR #145, opened 2017-03-21
Support for dynamic control of OpenThread NCP log level.

abtink:
This commit adds support for getting/setting a new wpan property
"OpenThread:LogLevel" tied to `SPINEL_PROP_DEBUG_NCP_LOG_LEVEL`
property. This property can be used to control the OpenThread NCP
log level dynamically.

By including this property in the `mSettings` hash table, it is
ensured that the user-selected log-level persists over NCP resets.

### DIFF
--- a/third_party/openthread/src/ncp/spinel.h
+++ b/third_party/openthread/src/ncp/spinel.h
@@ -227,6 +227,18 @@ enum
     SPINEL_MAC_PROMISCUOUS_MODE_FULL     = 2, ///< All decoded MAC packets are passed up the stack.
 };
 
+enum
+{
+    SPINEL_NCP_LOG_LEVEL_EMERG  = 0,
+    SPINEL_NCP_LOG_LEVEL_ALERT  = 1,
+    SPINEL_NCP_LOG_LEVEL_CRIT   = 2,
+    SPINEL_NCP_LOG_LEVEL_ERR    = 3,
+    SPINEL_NCP_LOG_LEVEL_WARN   = 4,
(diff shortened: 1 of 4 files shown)


third_party/openthread/src/ncp/spinel.h
```diff
@@ -227,6 +227,16 @@ enum
...
 
+enum
+{
+    SPINEL_NCP_LOG_LEVEL_NONE  = 0,
+    SPINEL_NCP_LOG_LEVEL_CRIT  = 1,
+    SPINEL_NCP_LOG_LEVEL_WARN  = 2,
+    SPINEL_NCP_LOG_LEVEL_INFO  = 3,
+    SPINEL_NCP_LOG_LEVEL_DEBG  = 4,
+    SPINEL_NCP_LOG_LEVEL_MAX   = SPINEL_NCP_LOG_LEVEL_DEBG,
+};
```
darconeous: Should we consider just using the [Syslog error-level enumeration](https://en.wikipedia.org/wiki/Syslog#Severity_level)?

abtink: The definition here maps to the current log levels defined in OT.

If we go to syslog enumeration then, we need to have a way to convert them to OT log level (which has fewer levels).

darconeous: That should be easy.

abtink:
The corresponding change in OT: https://github.com/openthread/openthread/pull/1492

PR #147, opened 2017-03-24
Refactor tunnel/netif-mgmt, differentiate UP and RUNNING, etc.

darconeous:
This commit moves the code related to address/route/linkstate management from `src/util/tunnel.[ch]` to the new `src/util/netif-mgmt.[ch]`. This abstraction is a better fit. This change also updates TunnelIPv6Interface accordingly.

One of the key changes in this commit is keeping track of the network interface link state. Doing `ifconfig wpan0 down` will now gracefully bring down the network interface. Likewise, doing `ifconfig wpan0 up` will bring it back online. (This isn't supported on macOS yet)

We now differentiate the `UP` and `RUNNING` flags internally, however the TUN driver on Linux does not make such a differentiation. Nonetheless, maintaining this differentiation may be useful in the future if we end up moving away from TUN.

The idea is that the `UP` flag is administratively controlled, whereas `RUNNING` is controlled by the driver. For example, the user can call `ifconfig eth0 up` to bring up their ethernet port, but if there is no cable plugged into it the interface won't have the `RUNNING` flag, and thus still isn't available. Likewise, the NCP may or may not be provisioned for a network. Internally, we treat this as another way to control the autoconnect property. If the interface is down, autoconnect is off. If the interface is up, autoconnect will be on.


darconeous:
ping @abtink

src/wpantund/NCPInstanceBase-NetInterface.cpp
```diff
@@ -34,6 +34,37 @@
...
+	if (isUp) {
+		set_property(kWPANTUNDProperty_DaemonAutoAssociateAfterReset, true);
+
+		if (!mEnabled) {
+			set_property(kWPANTUNDProperty_DaemonEnabled, true);
+		} else if (get_ncp_state() == COMMISSIONED) {
+			set_property(kWPANTUNDProperty_InterfaceUp, true);
+		}
+	} else {
+		set_property(kWPANTUNDProperty_DaemonAutoAssociateAfterReset, false);
```
abtink: Can this possibly cause a conflict with a user specified value for this property (say in the .conf file)?

darconeous: It is ideally just supposed to be another way that this variable can be set. We might just retire this property entirely.

src/util/netif-mgmt.c
```diff
@@ -0,0 +1,454 @@
...
+    struct in6_rtmsg rt;
+
+    memset(&rt, 0, sizeof(struct in6_rtmsg));
+
+	memcpy(rt.rtmsg_dst.s6_addr, route, sizeof(struct in6_addr));
+
+	rt.rtmsg_dst_len = prefixlen;
+    rt.rtmsg_flags = RTF_UP;
+    rt.rtmsg_metric = 512;
+
```
darconeous: Eww, will fix.

src/util/TunnelIPv6Interface.cpp
```diff
@@ -181,6 +203,19 @@ TunnelIPv6Interface::process(void)
...
+				bool isUp, isRunning;
+
+				if ((ifname == NULL) || (get_interface_name() != ifname)) {
+					continue;
+				}
+
+				isUp = ((ifinfo->ifi_flags & IFF_UP) == IFF_UP);
+				isRunning = ((ifinfo->ifi_flags & IFF_RUNNING) == IFF_RUNNING);
+
+				on_link_state_changed(isUp, isRunning);
```
abtink: The `on_link_state_changed()` seem to be called from a netlink change here. So if the driver itself brings up the interface we don't need to invoke the `on_link_state_changed()` and can rely on netlink event to get the callback, right?

src/wpantund/NCPInstanceBase-NetInterface.cpp
abtink: The property is definitely useful for debugging... let's keep it for now...

PR #150, opened 2017-04-06
ncp-spinel: Require CAP_POWER_SAVE to enter deep sleep via software.

darconeous:
This change avoids a potential churn loop that will occur if the interface is disabled when the NCP doesn't support `CAP_POWER_SAVE`.

In that case, we now simply turn off the stack, interface, and phy.

### DIFF
--- a/src/ncp-spinel/SpinelNCPTaskDeepSleep.cpp
+++ b/src/ncp-spinel/SpinelNCPTaskDeepSleep.cpp
@@ -86,22 +86,66 @@ nl::wpantund::SpinelNCPTaskDeepSleep::vprocess_event(int event, va_list args)
 			syslog(LOG_ERR, "DeepSleep: set_ncp_power(false) failed.");
 
 			// Turning off the power manually didn't work for some reason.
-			// Turn it back on and we will try to do it via the API.
+			// Turn it back on and we will try to do it via the API, below.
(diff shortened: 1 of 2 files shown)


src/ncp-spinel/SpinelNCPTaskDeepSleep.cpp
```diff
@@ -86,22 +86,69 @@ nl::wpantund::SpinelNCPTaskDeepSleep::vprocess_event(int event, va_list args)
...
+				SPINEL_PROP_NET_IF_UP,
+				false
+			);
+			EH_SPAWN(&mSubPT, vprocess_send_command(event, args));
+			check_noerr(mNextCommandRet);
+
+			// Turn off the phy
+			mNextCommand = SpinelPackData(
+				SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_BOOL_S),
+				SPINEL_PROP_PHY_ENABLED,
```
abtink: `SPINEL_PROP_PHY_ENABLED` is not settable in OT unless we have `OPENTHREAD_ENABLE_RAW_LINK_API`. 

I'd suggest we skip disabling the PHY and let OT control the radio. I think OpenThread will put the radio to sleep on `ThreadNetif::Down()`.

PR #151, opened 2017-04-07
update installation instructions for MacOS.

mszczodrak:
Address the change in brew --prefix vs brew --repository
Address the change in dbus installation path from d-bus to dbus


darconeous:
I donno, this contributor looks shady.... ;)

PR #154, opened 2017-04-22
Fix typo that prevented setting unsigned 64-bit properties

d-bahr:
Notably, this typo prevented setting XPANID as a 64-bit unsigned integer
instead of a byte array.


d-bahr:
Signed Google CLA.

darconeous:
I think you need to explicitly say `I signed it!` for @googlebot to pay attention.

darconeous:
Did not mean to close, those buttons are way too close!

d-bahr:
I signed it!

PR #158, opened 2017-04-29
Form network based on existing PANID and XPANID

### LINKED ISSUE #138: Make form/join parameter passing story consistent
After executing the following:
```
setprop Network:Key 00112233445566778899aabbccddeeff
setprop Network:PANID 0x8888
setprop Network:XPANID cafe00d00d00cafe
form -c 11 TEST
```
PANID and XPANID are not as specified, but rather randomly generated. SpinelNCPTaskForm checks the passed options for these values and if they are not passed - it generates new random ones. Tracing back to DBusIPCAPI_v0::interface_form_handler I can see that these values are not set there, which means that PANID and XPANID will always be overwritten.
However with network key it's slightly different - they are randomly generated only if they were not set before. IMO it should be like this with PANID and XPANID too.

Is this a bug or is there a reason for inconsistent behavior?

bukepo:
Setting PSKc relies on forming network with given XPANID. This PR fixes #138 .

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -162,7 +162,7 @@ nl::wpantund::peek_ncp_callback_status(int event, va_list args)
 }
 
 SpinelNCPInstance::SpinelNCPInstance(const Settings& settings) :
-	NCPInstanceBase(settings), mControlInterface(this)
+	NCPInstanceBase(settings), mControlInterface(this), mPanid(0xffff)
 {
 	mOutboundBufferLen = 0;
 	mInboundHeader = 0;
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -694,7 +694,6 @@ SpinelNCPInstance::set_property(
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_NetworkPANID)) {
 			uint16_t panid = any_to_int(value);
 
-			mCurrentNetworkInstance.panid = panid;
 			start_new_task(SpinelNCPTaskSendCommand::Factory(this)
 				.set_callback(cb)
 				.add_command(
@@ -821,7 +820,6 @@ SpinelNCPInstance::set_property(
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_NetworkXPANID)) {
 			Data xpanid = any_to_data(value);
 
-			memcpy(mCurrentNetworkInstance.hwaddr, xpanid.data(), xpanid.size());
 			start_new_task(SpinelNCPTaskSendCommand::Factory(this)
 				.set_callback(cb)
 				.add_command(
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -162,7 +162,7 @@ nl::wpantund::peek_ncp_callback_status(int event, va_list args)
 }
 
 SpinelNCPInstance::SpinelNCPInstance(const Settings& settings) :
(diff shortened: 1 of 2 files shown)


src/ncp-spinel/SpinelNCPTaskForm.cpp
```diff
@@ -41,17 +41,23 @@ nl::wpantund::SpinelNCPTaskForm::SpinelNCPTaskForm(
 ):	SpinelNCPTask(instance, cb), mOptions(options), mLastState(instance->get_ncp_state())
 {
 	if (!mOptions.count(kWPANTUNDProperty_NetworkPANID)) {
-		uint16_t panid;
+		uint16_t panid = instance->mPanid;
 
-		sec_random_fill(reinterpret_cast<uint8_t*>(&panid), sizeof(panid));
+		if (panid == 0xffff)
```
abtink: I think we need to initialize the `mPanId` to `0xffff` from constructor to make the un-specified panid in `form` generate the random value.

bukepo: Thanks, I've updated.

src/ncp-spinel/SpinelNCPInstance.h
```diff
@@ -241,6 +241,9 @@ class SpinelNCPInstance : public NCPInstanceBase {
 
 	int mSubPTIndex;
 
+	uint16_t mPanid;
+	uint64_t mXPanid;
```
darconeous: These are redundant. These fields are already present in `NCPInstanceBase:: mCurrentNetworkInstance`.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -694,7 +694,7 @@ SpinelNCPInstance::set_property(
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_NetworkPANID)) {
 			uint16_t panid = any_to_int(value);
 
-			mPanid = panid;
+			mCurrentNetworkInstance.panid = panid;
```
darconeous: This shouldn't be necessary, and doesn't follow the right model.

We set the PANID on the NCP, and then the NCP tells us that it set the PANID by telling us the new value of the PANID. We don't set the PANID or XPANID directly, we let the NCP tell us that these values have changed.

I'm willing to consider changes to that model with a good justification though.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -821,7 +821,7 @@ SpinelNCPInstance::set_property(
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_NetworkXPANID)) {
 			Data xpanid = any_to_data(value);
 
-			mXPanid = any_to_uint64(value);
+			memcpy(mCurrentNetworkInstance.hwaddr, xpanid.data(), xpanid.size());
```
darconeous: This is wrong for two reasons:

1. You are setting the xpanid to the wrong field. `hwaddr` is only used by scan results.
2. The same reason as my comment above for the PANID: The source of truth is the value of the property on the NCP, not in wpantund. We set the XPANID on the NCP and let the NCP tell us the XPANID in the response.

src/ncp-spinel/SpinelNCPTaskForm.cpp
```diff
@@ -52,7 +52,12 @@ nl::wpantund::SpinelNCPTaskForm::SpinelNCPTaskForm(
 	}
 
 	if (!mOptions.count(kWPANTUNDProperty_NetworkXPANID)) {
-		uint64_t xpanid = instance->mXPanid;
+		uint64_t xpanid;
+
+		memcpy(&xpanid, instance->mCurrentNetworkInstance.hwaddr, sizeof(xpanid));
```
darconeous: Using the wrong field again.

bukepo:
fixed, thanks for your suggestions. @darconeous @abtink

PR #160, opened 2017-04-30
Add border agent proxy features

bukepo:
This PR adds border agent proxy related properties and features.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -1487,6 +1487,7 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
 		uint16_t locator;
 		uint16_t port;
 		spinel_ssize_t ret;
+		Data data;
 
 		ret = spinel_datatype_unpack(
 			value_data_ptr,
@@ -1502,7 +1503,15 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
 
 		// Analyze the packet to determine if it should be dropped.
 		if ((ret > 0)) {
-			signal_border_agent_proxy_stream(frame_ptr, frame_len, locator, port);
+			// append frame
+			data.append(frame_ptr, frame_len);
+			// pack the locator in big endian.
+			data.push_back(locator >> 8);
+			data.push_back(locator & 0xff);
+			// pack the port in big endian.
+			data.push_back(port >> 8);
+			data.push_back(port & 0xff);
+			signal_property_changed(kWPANTUNDProperty_BorderAgentProxyStream, data);
 		}
 
 	} else if ((key == SPINEL_PROP_STREAM_NET) || (key == SPINEL_PROP_STREAM_NET_INSECURE)) {
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -300,6 +300,10 @@ SpinelNCPInstance::get_supported_property_keys()const
 		properties.insert(kWPANTUNDProperty_JamDetectionDebugHistoryBitmap);
 	}
 
+	if (mCapabilities.count(SPINEL_CAP_THREAD_BA_PROXY)) {
+		properties.insert(kWPANTUNDProperty_BorderAgentProxyEnabled);
+	}
+
 	if (mCapabilities.count(SPINEL_CAP_NEST_LEGACY_INTERFACE))
 	{
 		properties.insert(kWPANTUNDProperty_NestLabs_LegacyMeshLocalPrefix);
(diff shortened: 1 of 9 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -959,6 +983,24 @@ SpinelNCPInstance::set_property(
 				.finish()
 			);
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_BorderAgentProxyStream)) {
+			Data packet = any_to_data(value);
+
+			uint16_t port = *(uint16_t*)(packet.data() + packet.size() - sizeof(port));
```
abtink: The pointer type-cast model used here is not really safe to memory alignments. 
Better to read each bytes from packet/array and then convert the values to the type.  We also need to be careful of endian-ness of host (e.g., how `uint16` is read from memory may be different on different platforms)...

bukepo: Yes, this is encoded in network byte order, as you see in the next line, it converts it into host byte border. I will update the implementation here to make it more clear.

src/ipc-dbus/wpantund.conf
```diff
@@ -28,6 +28,10 @@
         <allow send_interface="com.nestlabs.WPANTunnelDriver.PermitJoin"/>
         <allow send_interface="com.nestlabs.WPANTunnelDriver.Status"/>
         <allow send_interface="com.nestlabs.WPANTunnelDriver.List"/>
+
+        <!-- Border Agent -->
+        <allow own="ot.ba"/>
```
darconeous: We should make this a real reverse DNS name. Not sure what reverse DNS name we should use, though. We are currently using `org.wpantund.*`, which would be appropriate if this is tied to wpantund.

darconeous: Oh, I see what you are doing. Sorry I misunderstood. `wpantund` talks to the border agent using this API, this API isn't hosted on `wpantund`.

How does the border agent send unsolicited packets back?

I wonder if we should change the approach to be that the border agent checks in with wpantund and then uses D-Bus signals to indicate received packets, and the border agent simply calls a method to send a packet.

Alternatively, you could implement this feature entirely using properties, without developing any new D-Bus protocol. This would work by creating a new property that would act similarly to the associated spinel property. You'd set it to send packets, and you would receive change notifications for inbound packets.

Thoughts?

darconeous:
Oh, I see what you are doing. Sorry I misunderstood. `wpantund` talks to the border agent using this API, this API isn't hosted on `wpantund`.

How does the border agent send unsolicited packets back?

I wonder if we should change the approach to be that the border agent checks in with wpantund and then uses D-Bus signals to indicate received packets, and the border agent simply calls a method to send a packet.

Alternatively, you could implement this feature entirely using properties, without developing any new D-Bus protocol. This would work by creating a new property that would act similarly to the associated spinel property. You'd set it to send packets, and you would receive change notifications for inbound packets.

Thoughts?

bukepo:
@darconeous thanks for your suggestions. The current implementation is sending packets to NCP by **setting the property** `BorderAgentProxy:Stream`.
I also tried to use existing mechanism for sending packets to border agent. However it's 1:1 communication, not 1:N communication. Using signal would cause multiple border agents receiving the same packet, if there happened to be multiple border agents being started. Any thoughts?

darconeous:
If there is more than one border agent running, I think it would be pretty simple to make sure that each border agent tries to register itself with D-Bus using a name that includes the network interface name. That way only one instance of the border agent would be allowed per network interface name.

I'm not too worried about the broadcast nature of signals. Signals only get routed to processes that have expressed interest in those signals. In the case where there is more than one border agent for a given interface I'd consider that a configuration error---one easily prevented using the above trick.

bukepo:
thanks, I will update as you suggested.

bukepo:
I have updated as suggested. Thanks. @darconeous

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -1439,6 +1481,30 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
...
+			&frame_len,
+			&locator,
+			&port
+		);
+
+		__ASSERT_MACROS_check(ret > 0);
+
+		// Analyze the packet to determine if it should be dropped.
+		if ((ret > 0)) {
+			signal_border_agent_proxy_stream(frame_ptr, frame_len, locator, port);
```
darconeous: Why are we making a special case? Why not just use the existing property changed mechanism instead of doing all of this new plumbing?

darconeous:
Thanks for bearing with me, I think we are really close!

src/ncp-spinel/SpinelNCPInstance.cpp
bukepo: Do you mean I should encode `frame`, `locator` and `port` into one `Data` type value?

darconeous: Yes, similar to how we do for spinel. I would personally put the locator and port as the first four bytes, and then have the coap frame be the remaining bytes.

bukepo: Thanks for your suggestions, I will update soon.

bukepo:
@darconeous I've updated the code, and I put the locator and port at the end to reduce the cost for resizing. Thanks for your suggestions.

PR #161, opened 2017-05-03
Add Network:Role and Thread:DeviceMode properties.

d-bahr:
Network:Role allows changing between router, child, leader, and detached, similar to the CLI "state" command.

Thread:DeviceMode allows changing the device functionality similar to the CLI "mode" command.


darconeous:
Ugg, that's what I get for fixing the commit directly in GitHub (GoogleBot doesn't recognize my personal email address). In this case we can ignore it.

darconeous:
### Note to self: Don't "Merge", do a "Squash and Merge".

PR #163, opened 2017-05-10
tunnel: Allow for using either UTUN or TunTapOSX on macOS

darconeous:
This commit allows the user to use either the native UTUN interface or the TunTapOSX third-party driver for the network tunnel on macOS.

This is useful because it allows wpantund to be started as non-root on macOS, by changing the Unix access permissions on the `/dev/tunX` files.

### DIFF
--- a/src/util/tunnel.c
+++ b/src/util/tunnel.c
@@ -43,7 +43,7 @@
 #include <sys/ioctl.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


darconeous:
Note that this change hasn't yet been tested.

abtink:
Closing older stale PR.

PR #164, opened 2017-05-11
Update `spinel.h` and `spinel.c` from OpenThread

abtink:
This commit updates the spinel files from OpenThread. This syncs
the files with OpenThread at commit 1ece777a8232.

### DIFF
--- a/third_party/openthread/src/ncp/spinel.c
+++ b/third_party/openthread/src/ncp/spinel.c
@@ -1199,6 +1199,22 @@ spinel_prop_key_to_cstr(spinel_prop_key_t prop_key)
         ret = "PROP_THREAD_CONTEXT_REUSE_DELAY";
         break;
 
+    case SPINEL_PROP_THREAD_DISCOVERY_SCAN_JOINER_FLAG:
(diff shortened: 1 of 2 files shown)


third_party/openthread/src/ncp/spinel.c
```diff
@@ -1199,6 +1199,22 @@ spinel_prop_key_to_cstr(spinel_prop_key_t prop_key)
         ret = "PROP_THREAD_CONTEXT_REUSE_DELAY";
         break;
 
+    case SPINEL_PROP_THREAD_DISCOVERY_SCAN_JOINER_FLAG:
+        ret = "SPINEL_PROP_THREAD_DISCOVERY_SCAN_JOINER_FLAG";
```
darconeous: These should return just the official name of the property from the Spinel spec, which doesn't include the `SPINEL_` prefix. The `SPINEL_` prefix is just for constants in `spinel.h`.

I won't hold back approval on this though.

abtink: Good catch... Will fix it on OT side...

PR #166, opened 2017-05-15
Add support for controlling steering data when joinable

abtink:
This commit adds new feature to allow setting of the steering data
when the joinability status gets changed (through `permit_join()`).
This feature requires the NCP `SPINEL_CAP_OOB_STEERING_DATA`
capability.

This commit defines two new wpan properties:

1) `OpenThread:SteeringData:SetWhenJoinable` (boolean) which defines
whether or not when the device is joinable the steering data is set.
The default setting is `false`. It is recommended that this property
is changed using `wpantund.conf` during initialization.

2) `OpenThread:SteeringData:Address` (eui64 value) which specifies the
address used to set the steering data (out of band) using the spinel
property `SPINEL_PROP_THREAD_STEERING_DATA` (the given eui64 is used
to compute the bloom filter which is used as the steering data).
Default value is all `0xff`s which means set the steering data/bloom
filter to accept/allow all.

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -404,11 +404,20 @@ SpinelNCPControlInterface::permit_join(
     )
 {
 	SpinelNCPTaskSendCommand::Factory factory(mNCPInstance);
+	bool should_update_steering_data = false;
+	uint8_t steering_data_addr[sizeof(mNCPInstance->mSteeringDataAddress)];
+
 	int ret = kWPANTUNDStatus_Ok;
 
 	if (!mNCPInstance->mEnabled) {
 		ret = kWPANTUNDStatus_InvalidWhenDisabled;
 		goto bail;
+	}
+
+	if (mNCPInstance->mCapabilities.count(SPINEL_CAP_OOB_STEERING_DATA)
+		&& mNCPInstance->mSetSteeringDataWhenJoinable
+	) {
+		should_update_steering_data = true;
 	}
 
 	if (traffic_port == 0) {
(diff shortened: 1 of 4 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -1036,6 +1045,22 @@ SpinelNCPInstance::set_property(
 				.finish()
 			);
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_OpenThreadSteeringDataSetWhenJoinable)) {
+			mSetSteeringDataWhenJoinable = any_to_bool(value);
+			cb(kWPANTUNDStatus_Ok);
+
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_OpenThreadSteeringDataAddress)) {
```
darconeous: How do you clear this out? Just set it to all zeros?

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -429,11 +438,25 @@ SpinelNCPControlInterface::permit_join(
...
+		memcpy(steering_data_addr, mNCPInstance->mSteeringDataAddress, sizeof(steering_data_addr));
+
 	} else {
+
 		factory.add_command(SpinelPackData(
 			SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_NULL_S),
 			SPINEL_PROP_THREAD_ASSISTING_PORTS
 		));
+
+		memset(steering_data_addr, 0, sizeof(steering_data_addr));
```
darconeous: Should `should_update_steering_data` be set to true here? Or perhaps explicitly add another command which sets an empty payload for `SPINEL_PROP_THREAD_STEERING_DATA `?

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: Yes. This mirrors the open thread API behavior

src/ncp-spinel/SpinelNCPControlInterface.cpp
abtink: `should_update_steering_data` is updated in `if` before this (it checks if the capability is present and we have `mSetSteeringDataWhenJoinable` set). 

Here we only decide what to set it to (either all zero, or the value `mSteeringDataAddress`)...

PR #169, opened 2017-05-16
improve the config-gateway function

LeZhang2016:
- add a property named "Thread:OnMeshPrefixes" which indicates the information about gateway (prefix, flags, and priority).
- add "priority" for configured prefix.
- provide "flags" (preferred, slaac, onMesh) for OT border router setting

### DIFF
--- a/src/ipc-dbus/DBusIPCAPI_v1.cpp
+++ b/src/ipc-dbus/DBusIPCAPI_v1.cpp
@@ -1192,7 +1192,7 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
 	int prefix_len_in_bytes(0);
 	struct in6_addr address = {};
 	bool did_succeed(false);
-	int16_t priority_raw;
+	int16_t priority_raw(0);
 	NCPControlInterface::OnMeshPrefixPriority priority(NCPControlInterface::PREFIX_MEDIUM_PREFERENCE);
 
 	did_succeed = dbus_message_get_args(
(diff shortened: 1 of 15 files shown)


darconeous:
I think this commit is missing some files you added.

LeZhang2016:
@darconeous, Thanks a lot for your reminder. I have fixed it.

src/ipc-dbus/DBusIPCAPI_v1.cpp
```diff
@@ -1028,19 +1028,28 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
...
+	int16_t priority_raw;
+	NCPControlInterface::OnMeshPrefixPriority priority(NCPControlInterface::PREFIX_MEDIUM_PREFERENCE);
 
 	did_succeed = dbus_message_get_args(
 		message, NULL,
 		DBUS_TYPE_BOOLEAN, &default_route,
 		DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &prefix, &prefix_len_in_bytes,
 		DBUS_TYPE_UINT32, &preferred_lifetime,
 		DBUS_TYPE_UINT32, &valid_lifetime,
+		DBUS_TYPE_BOOLEAN, &preferred,
```
abtink: This changes the dbus message format and  breaks compatibility with sw using existing version (using existing dbus message format).

We should check for the case where the new parameters are not provided, which should cause a `did_succeed` failure, and then parse it up to old parameters (use default values for new ones).  `DBusIPCAPI_v1::interface_route_add_handler()` is an example of how it's done for another command.

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -435,6 +437,19 @@ NCPInstanceBase::get_property(
 	} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_NetworkNodeType)) {
 		cb(0, boost::any(node_type_to_string(mNodeType)));
 
+	} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_ThreadOnMeshPrefixes)) {
+		std::list<std::string> result;
+		std::map<struct in6_addr, GlobalAddressEntry>::const_iterator it;
+		static const char flag_lookup[] = "ppPSDCRM";
+		char address_string[INET6_ADDRSTRLEN];
+		for ( it = mOnMeshPrefixes.begin();
+			  it != mOnMeshPrefixes.end();
```
abtink: nit: The alignment/spacing for this `for`.

src/ipc-dbus/DBusIPCAPI_v1.cpp
LeZhang2016: @abtink, Thanks a lot for your comments, I will modify this part according to your suggestion.

src/wpantund/NCPInstanceBase.cpp
LeZhang2016: Thanks, I will modify it.

abtink:
Looks great. Thanks for doing this.

PR #174, opened 2017-05-26
Add support for MLE discover scan

abtink:
This commit includes the following changes: It updates the `TaskScan`
implementation to add support for the discovery scan in addition
energy and network scan operations It also adds a new dbus method for
MLE discovery scan. The wpanctl "scan" command is  updated to add
options for the new discover scan ("scan -d" to start a discovery
scan).


abtink:
`WPANTUND_IF_CMD_DISCOVER_SCAN_START` is the (new) dbus command which contains all the options for MLE discovery. 

We can consider combining NET_SCAN and DISCOVER_SCAN into a single dbus command say NET_SCAN_2 but I don't see any immediate advantage for it. I think keeping the two commands separate makes it easier to use them now.

Also I always felt that the two operations (traditional network scan and MLE discover scan) are from a network-perspective sort of different: one should be available/supported on any 15.4-capable radio (and is a MAC level operation), the other one is very Thread specific (and uses Thread MLE layer).

PR #177, opened 2017-06-01
Fix some compiling warnings

bukepo:
This PR simply fixes some annoying warnings.

### DIFF
--- a/src/wpanctl/tool-cmd-commissioner.c
+++ b/src/wpanctl/tool-cmd-commissioner.c
@@ -305,7 +305,7 @@ int tool_cmd_commissioner(int argc, char* argv[])
 					goto bail;
 				}
 				if (strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1) != EXT_ADDRESS_LENGTH_CHAR) {
-					fprintf(stderr, "%s: error: Wrong address length.%d \n", argv[0], strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1));
+					fprintf(stderr, "%s: error: Wrong address length.%zu \n", argv[0], strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1));
 					ret = ERRORCODE_BADARG;
 					goto bail;
 				}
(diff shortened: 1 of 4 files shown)


src/wpanctl/tool-cmd-commissioner.c
```diff
@@ -305,7 +305,7 @@ int tool_cmd_commissioner(int argc, char* argv[])
 					goto bail;
 				}
 				if (strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1) != EXT_ADDRESS_LENGTH_CHAR) {
-					fprintf(stderr, "%s: error: Wrong address length.%d \n", argv[0], strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1));
+					fprintf(stderr, "%s: error: Wrong address length.%zu \n", argv[0], strnlen(ext_addr, EXT_ADDRESS_LENGTH_CHAR+1));
```
darconeous: For the sake of portability, I'd recommend keeping `%d` and just casting the `strlen` to `int`.

bukepo: >OpenThread uses and enforces the ISO9899:1999 (aka ISO C99, C99) C language standard as the minimum.

What does wpantund uses and enforces? **z** is in standard C99.

darconeous: The problem is that `strnlen` returns a `size_t`, which may be a `unsigned long long` or an `unsigned long` or even an `unsigned int`, depending on the platform. This means that there isn't a good universal way to handle the situation without casting.

bukepo: Is `z` modifier designed to solve this portability issue? If it's not allowed, shall we make it clear in `CONTRIBUTING.md` or other place that we should avoid new features introduced in C99?
I think other contributor may also confused about what to use and what not.

darconeous: Eh, C99 is fine. I guess my old pre-C99 instincts are kicking in. This is fine.

PR #183, opened 2017-06-07
Adding new wpan property "Thread:OffMeshRoutes" and its 'get' support

abtink:
The wpan property is related to `SPINEL_PROP_THREAD_OFF_MESH_ROUTES`
property.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -252,6 +252,7 @@ SpinelNCPInstance::get_supported_property_keys()const
 		properties.insert(kWPANTUNDProperty_ThreadChildTable);
 		properties.insert(kWPANTUNDProperty_ThreadNeighborTable);
 		properties.insert(kWPANTUNDProperty_ThreadCommissionerEnabled);
+		properties.insert(kWPANTUNDProperty_ThreadOffMeshRoutes);
 	}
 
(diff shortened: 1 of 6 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -380,6 +381,80 @@ static int unpack_jam_detect_history_bitmap(const uint8_t *data_in, spinel_size_
...
+			&route_prefix,
+			&prefix_len,
+			&is_stable,
+			&flags,
+			&is_local,
+			&next_hop_is_this_device
+		);
+
+		if (len <= 0) {
+			ret = kWPANTUNDStatus_Failure;
```
darconeous: Is there a more specific error we could use here?

abtink: Any suggestions?

abtink: We seems to use this error type from other tasks in case of a spinel parsing/unpacking error (e.g., in `SpinelNCPTaskSendCommand` for type parser which is basically used for all `get_prop`).

darconeous: Ah, yes, it seems that there isn't one specifically for parse failure. Perhaps we should add one, but we can leave this as is for now.

abtink: I agree. We should later add a `parse` wpan error type and use it for this (matches well with spinel one).

PR #184, opened 2017-06-08
IP Counters

aeliot:
This is the companion PR to https://github.com/openthread/openthread/pull/1892

Adds in support for IP level counters.


third_party/openthread/src/ncp/spinel.c
```diff
@@ -1479,6 +1479,22 @@ spinel_prop_key_to_cstr(spinel_prop_key_t prop_key)
         ret = "PROP_CNTR_RX_SPINEL_OUT_OF_ORDER_TID";
         break;
 
+    case SPINEL_PROP_CNTR_IP_TX_SUCCESS:
+        ret = "SPINEL_PROP_CNTR_IP_TX_SUCCESS";
```
darconeous: You need to remove the `SPINEL_` prefix in the returned string.

aeliot: I made the chance in wpantund and openthread

PR #185, opened 2017-06-09
Notes about Docker

jaredwolff:
I spent some cycles trying to figure out if the Docker setup in /etc/ would have been useful for my hardware testing. Unfortunately not (at least on OS X) I added a Readme so people don't waste their time doing the same. Also the `build-in-docker.sh` was out of date and was not consistent with the other documentation. 

Summary:

Adding notes about the end-to-end process of using Docker.
Updating the `build-in-docker.sh` to run the bootstrap script before configure

### DIFF
--- a/etc/Readme.md
+++ b/etc/Readme.md
@@ -1,10 +1,29 @@
 # Getting started with Docker
 
-1. Install Docker
-1. Run
+1. [Install Docker](https://docs.docker.com/engine/installation/)
+1. Build the Image (from the etc folder)
 
-    ./etc//build-in-docker.sh
+      cd etc
+      docker build -t wpantund .
 
-**Note:** for OSX users, you **will not** be able to share your serial device with the 
+1. Run (from base directory)
+
+      cd ..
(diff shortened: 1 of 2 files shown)


jaredwolff:
Trying to fix the CLA error. Used my website email hello@jaredwolff.com for the commits.

darconeous:
First, thanks for giving this some attention! I agree this needs to be better documented. I implemented it to help me do quick linux-build sanity checks on changes before making upstream pushes and never got around to documenting it.

The intended workflow was to  (if necessary) call the `./bootstrap.sh` script separately and to then call either `etc/build-in-docker.sh` (for a generic linux build test) or `etc/run-in-docker.sh` (for performing specific actions). If the docker image doesn't exist, the `run-in-docker.sh` script automatically builds it—so you don't actually have to directly invoke `docker` at all.

I'll likely use your work here as a starting point for documenting these details.

abtink:
Closing older stale PR.

PR #186, opened 2017-06-09
Add support for getting/setting "NCP:TxPower" property

abtink:


darconeous:
I approve, but I wonder if we should store the set value in process so that we can restore the value after a reset, like we do for the TMSP plugin?

darconeous:
That can be added as a separate pull request.

abtink:
@darconeous  we save the tx power value when from the property setter in the `mSettings` table. So if NCP is reset we will set all the saved values back on it.

```
    mSettings[kWPANTUNDProperty_NCPTXPower] = SettingsEntry(command);
```

PR #190, opened 2017-06-14
speeding up building

bukepo:
Currently building wpantund takes a long time, especially on some arm boards. The main cause is boost library, which utilizes a lot of templates. One heavy component is the `boost/signals2` component.

Wpantund uses the `boost/signals2` to bubble events in several components, which results in headers introduced by `boost/signals2` to be preprocessed several times. This PR collects these signals into one file `src/util/Signals.h` so that headers of `boost/signals2` will only be processed once.
I'm not sure if this is the right solution to this problem, but here's the time usage on i386 single core.

```sh
time make
```
### before
real    2m34.542s
user    2m24.156s
sys     0m8.164s

### after
real    1m34.206s
user    1m27.180s
sys     0m4.972s


darconeous:
Ultimately, this is a worthwhile goal. The time it takes to build wpantund is quickly becoming a top complaint. We discussed your changes here earlier, and while I share your reservations on this approach, I think it is a good conversation starter.

It looks like both [Clang](https://clang.llvm.org/docs/PCHInternals.html) and [GCC](https://gcc.gnu.org/onlinedocs/gcc/Precompiled-Headers.html) support pre-compiled headers. Perhaps that might be a good place to invest effort that will lead to a minimum of structural changes.

[`ccache`](https://ccache.samba.org/] may also be helpful in reducing rebuild times (as opposed to raw compile times), but I'm not sure if that is something that should be built into our build system or something that should be explicitly invoked by the user (by setting `CC="ccache clang"` and `CXX="ccache clang++"` on the `configure` command line, for example).

darconeous:
I've managed to get the build time on my MacBook Pro down from 2m51s to 1m45s using Clang's pre-compiled header support. I'll file a pull request once I get it cleaned up.

darconeous:
Have a look at https://github.com/openthread/wpantund/pull/193/files

abtink:
@bukepo , I am closing the older PRs on wpantund. Feel free to re-submit this. Thanks

PR #191, opened 2017-06-14
Initial framework for fuzz testing

darconeous:
This is the first step in implementing a comprehensive fuzz test target
for wpantund. At the moment, the target is only fuzzing the config
file parser, which is nice but ultimately not really what we are interested
in. The plan is to extend this fuzz target to cover a much wider set
of input processing.

This fuzz target is currently set up for use by [libFuzzer][1], but
may in the future be used by other fuzzing systems.

Additional fuzz targets will be set up for more low-level module fuzzing.

[1]: http://llvm.org/docs/LibFuzzer.html


darconeous:
@bukepo: Did you ever get a chance to work on adding the NCP fuzzing like we discussed when I was in Shanghai?

bukepo:
I'm sorry but I'm still engaged in some urgent tasks. I will work on this once I got the time.

darconeous:
FYI: See the [Fuzzing page](https://github.com/openthread/wpantund/wiki/Fuzzing) on the wiki for instructions on how to fuzz.

PR #193, opened 2017-06-16
build: Precompiled Header Support

darconeous:
This commit adds support for both GCC and Clang precompiled headers. Precompiled headers are used automatically if GCC or Clang is detected, but it may be explicitly disabled by passing the `--disable-pch` argument to the configure script.

The strategy used is to compute a precompiled header for each target and include it when compiling each file. This should significantly improve build times.

Note that when used with GCC, the current default is to NOT use precompiled headers. This is because we don't seem to notice any significant build performance improvement (quite the opposite, in fact) with the current PCH strategy with GCC. You can explicitly enable PCH with GCC by setting `--enable-pch` on the configure script command line.


darconeous:
This change didn't have any noticeable impact on the macOS Travis build, which is suspicious. I'm investigating.

abtink:
Closing older PR.

PR #194, opened 2017-06-16
Ensure to switch to ASSOCIATED on NCP role change from detached

abtink:
This change fixes an issue where we could remain in the ISOLATED
state even after receiving an NCP event with role change to other
than detached.

Here is the flow to get stuck in isloated state
- A NCP role change to detached will trigger a `change_ncp_state(ISOLATED);`
- Later if we get a role change back to router, we won't change the wpantund state
- The change uses `ncp_state_is_joining_or_joined()` which includes the `ISOLATED` state to update.


abtink:
This is verified/reviewed internally. Merging the change.

PR #196, opened 2017-06-26
Rx/Tx total time stats

aeliot:
This PR adds tracking for the amount of time the radio spends in Rx or Tx state.

### DIFF
--- a/third_party/openthread/src/ncp/spinel.h
+++ b/third_party/openthread/src/ncp/spinel.h
@@ -633,6 +633,24 @@ typedef enum
     SPINEL_PROP_JAM_DETECT_HISTORY_BITMAP
                                         = SPINEL_PROP_PHY_EXT__BEGIN + 5,
 
+    /// Total Tx Time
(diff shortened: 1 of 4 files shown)


third_party/openthread/src/ncp/spinel.h
```diff
@@ -633,6 +633,24 @@ typedef enum
...
                                         = SPINEL_PROP_PHY_EXT__BEGIN + 5,
 
+    /// Total Tx Time
+    /** Fromat: `L` (read-only)
+     *  Units: ms
+     *
+     * This property returns the totaly amount of time the radio has been
+     * in Tx mode.
+     */
+    SPINEL_PROP_TX_TOTAL_TIME           = SPINEL_PROP_PHY_EXT__BEGIN + 6,
```
darconeous: What are the units?

And spinel properties should always be prefixed with their section, like `PROP_PHY_TX_TOTAL_TIME`...

PR #208, opened 2017-07-28
Allow controlling of steering data independent of joinability state

abtink:
This commit allows full control of steering data independent of
joinability status of device. Changing steering data requires
`SPINEL_CAP_OOB_STEERING_DATA` capability on NCP. The wpan property
`OpenThread:SteeringData:Address`  (EUI64 value) can be used to
directly change steering data on NCP (using the spinel property
`SPINEL_PROP_THREAD_STEERING_DATA`). Default value of steering data is
all zeros (which means there is no steering data (no bloom filter)).

This commit also ensures to retain the last value of steering data in
`mSettings` list so that it is re-applied on an NCP reset.

This commit also removes `OpenThread:SteeringData:SetWhenJoinable`
(boolean) wpantund property and the code associated with it.

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -417,20 +417,11 @@ SpinelNCPControlInterface::permit_join(
 	)
 {
 	SpinelNCPTaskSendCommand::Factory factory(mNCPInstance);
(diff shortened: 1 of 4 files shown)


abtink:
After discussion deiced to enhance this feature to add more control of the behavior. Closing this PR for now.

PR #209, opened 2017-08-03
IPv6 address (unicast/multicast) and prefix entries

abtink:

It adds support for getting/updating multicast IPv6 address table from
NCP using spinel property `SPINEL_PROP_IPV6_MULTICAST_ADDRESS_TABLE`.

It also changes the model for tracking of addresses/prefixes.
`NCPInstanceBase` class keeps track of unicast/multicast and on-mesh-
prefixes in different tables. When an entry is added, the origin of the
entry is also recorded. We have three possible origins: (1) thread
NCP, (2) primary (tunnel) interface, or (3) user through a direct
command. The `NCPInstanceBase` is notified about updates to
addresses/prefixes through a set of methods which are then used by
tunnel interface and NCP plug-in implementation. If there is a change
`NCPInstanceBase` ensures to update other entities with newly
added/removed entry (e.g., if an address is added on the primary
interface, it is then updated on the NCP and vice versa). The new model
ensures that entries added by user (or through the primary interface)
persist over NCP reset. It also disallows an address/prefix added by
user to be removed by NCP or other entities.

Another change in this commit is on how on-mesh-prefixes corresponding
to new addresses are added. When a unicast IPv6 address is added on
primary interface, a corresponding prefix is derived and applied to
the NCP. A new wpan property "Daemon:SetDefaultRouteForAutoAddedPrefix"
is added to control the flags set for prefixes derived from an added
unicast address. If set to `true` when a prefix is auto-added it will
include the default route flag, otherwise it is set with this flag
cleared. The default setting for this property is false. The process
of adding SLACC addresses based on prefixes is also now done by the
`NCPInstanceBase`.

The `TunnelIPv6Interface` class is also modified to add support for
joining/leaving multicast addresses on interface. Another change is on
how the addresses are updated on the interface. If the interface is
offline, the addresses added to `TunelIpV6Interface` (using its
methods) are kept in a table and when the interface is brought online
the addresses are added. The new implementation also waits to get a
confirmation (through netlink) that an address is added before removing
the table entry (this helps with situation where adding an address may
first remove it on interface thus triggering an extra removed callback
notification).


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -605,17 +598,14 @@ unpack_thread_off_mesh_routes(const uint8_t *data_in, spinel_size_t data_len, bo
...
 			NCPControlInterface::ExternalRoutePriority priority;
 
 			priority = SpinelNCPControlInterface::convert_flags_to_external_route_priority(flags);
 
-			inet_ntop(AF_INET6,	route_prefix, address_string, sizeof(address_string));
-
 			snprintf(c_string, sizeof(c_string),
 				"%s/%d, stable:%s, local:%s, next_hop:%s, priority:%s (flags:0x%02x)",
-				address_string,
+				in6_addr_to_string(*route_prefix).c_str(),
```
darconeous: Somewhat unrelated comment:

I realize we are doing this in a lot of places, but I'll admit I'm not entirely sure what the scope of the returned `std::string` object is. If the scope is the evaluation of this argument, then all lines like this represent a use-after-free error. However, if the scope of evaluation is the entire line, then it is perfectly acceptable. I'm 90% sure it is the later case because I have trouble imagining such a condition not causing obvious problems with lots of code. But this is one of those areas where my C++ knowledge comes up a tad short.

Since we are using this pattern all over the place, I don't consider this to be blocking of this change. This is just something that we should confirm the behavior of.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2556,93 +2539,183 @@ SpinelNCPInstance::handle_ncp_spinel_callback(unsigned int command, const uint8_
 	process_event(EVENT_NCP(command), cmd_data_ptr[0], cmd_data_ptr, cmd_data_len);
 }
 
-void
-SpinelNCPInstance::address_was_added(const struct in6_addr& addr, int prefix_len)
+bool
+SpinelNCPInstance::should_filter_address(const struct in6_addr &addr, uint8_t prefix_len)
```
darconeous: Shouldn't we be filtering out the RLOC addresses at the NCP?

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2556,93 +2539,183 @@ SpinelNCPInstance::handle_ncp_spinel_callback(unsigned int command, const uint8_
...
+SpinelNCPInstance::remove_unicast_address_on_ncp(const struct in6_addr& addr, uint8_t prefix_len, CallbackWithStatus cb)
 {
-	if (status == kWPANTUNDStatus_Timeout)
-	{
-		syslog(LOG_ERR, "Timed out while performing \"%s\" - Resetting NCP.", operation.c_str());
-		ncp_is_misbehaving();
-	}
+	SpinelNCPTaskSendCommand::Factory factory(this);
+
+	syslog(LOG_NOTICE, "Removing address \"%s/%d\" from NCP", in6_addr_to_string(addr).c_str(), prefix_len);
```
darconeous: I'd personally consider this `LOG_INFO`, rather than `LOG_NOTICE`.

src/util/TunnelIPv6Interface.cpp
```diff
@@ -103,24 +103,79 @@ TunnelIPv6Interface::on_link_state_changed(bool isUp, bool isRunning)
...
+					continue;
+				}
+
+				syslog(LOG_INFO, "Joining multicast address \"%s\" on interface \"%s\".",
+				       in6_addr_to_string(iter->first).c_str(), mInterfaceName.c_str());
+
+				IGNORE_RETURN_VALUE(netif_mgmt_join_ipv6_multicast_address(mNetifMgmtFD, mInterfaceName.c_str(),
+				                    iter->first.s6_addr));
+			}
+			mMulticastAddresses.clear();
```
darconeous: ?

src/util/TunnelIPv6Interface.cpp
```diff
@@ -103,24 +103,79 @@ TunnelIPv6Interface::on_link_state_changed(bool isUp, bool isRunning)
...
 }
 
+void
+TunnelIPv6Interface::on_address_added(const struct in6_addr &address, uint8_t prefix_len)
+{
+	if (mUnicastAddresses.count(address)) {
+		mUnicastAddresses.erase(address);
+	}
+
+	syslog(LOG_NOTICE, "TunnelIPv6Interface: \"%s/%d\" was added to \"%s\"", in6_addr_to_string(address).c_str(), prefix_len,
```
darconeous: I personally think this line (and all other similar "address added/removed" log lines) should be `LOG_INFO` or (in some cases) `LOG_DEBUG`, IMHO.

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -30,122 +30,529 @@
...
+}
+
+void
+NCPInstanceBase::UnicastAddressEntry::set_valid_lifetime(uint32_t valid_lifetime)
+{
+	mValidLifetime = valid_lifetime;
+
+	mValidLifetimeExpiration = (valid_lifetime == UINT32_MAX)
+		? TIME_DISTANT_FUTURE
+		: time_get_monotonic() + valid_lifetime;
```
darconeous: This is tricky, because last time I checked this sort of thing wasn't happening on the NCP. What action do we take when this expires? I would think the NCP would be in charge of expiring things.

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -30,122 +30,529 @@
...
+	}
+
+	return std::string(c_string);
+}
+
+// ========================================================================
+// MARK: Global Entries Management
 
+bool
+NCPInstanceBase::should_filter_address(const struct in6_addr &address, uint8_t prefix_len)
```
darconeous: Can you elaborate on the use cases for this?

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -583,6 +587,13 @@ NCPInstanceBase::property_set_value(
 				reinitialize_ncp();
 			}
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_DaemonSetDefRouteForAutoAddedPrefix)) {
+			mSetDefaultRouteForAutoAddedPrefix = any_to_bool(value);
+			cb(0);
+			if (mTerminateOnFault && (get_ncp_state() == FAULT)) {
+				reinitialize_ncp();
```
darconeous: I don't understand the logic here with `mTerminateOnFault`. How would we ever get to this code?

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -843,6 +854,7 @@ NCPInstanceBase::handle_ncp_state_change(NCPState new_ncp_state, NCPState old_nc
 	// (Special case of InterfaceUp -> InterfaceDown)
 	} else if (ncp_state_is_commissioned(old_ncp_state)
 		&& !ncp_state_is_commissioned(new_ncp_state)
+		&& !ncp_state_is_joining(new_ncp_state)
```
darconeous: Can you elaborate a little bit on this change? It seems unrelated to the rest.

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -854,7 +866,6 @@ NCPInstanceBase::handle_ncp_state_change(NCPState new_ncp_state, NCPState old_nc
 	// and we have global addresses, then need to clear them out.
 	} else if (old_ncp_state == UNINITIALIZED
 		&& new_ncp_state == OFFLINE
-		&& !mGlobalAddresses.empty()
```
darconeous: If you are going to remove this line, you should also change the above comment.

src/wpantund/NCPInstanceBase.h
```diff
@@ -136,38 +136,71 @@ class NCPInstanceBase : public NCPInstance, public EventHandler{
...
-	// ========================================================================
-	// MARK: Global Address Management
+	void unicast_address_was_added(Origin origin, const struct in6_addr &address,
+			uint8_t prefix_len = 64, uint32_t valid_lifetime = UINT32_MAX, uint32_t preferred_lifetime = UINT32_MAX);
 
+	void unicast_address_was_removed(Origin origin, const struct in6_addr &address);
 
-	void add_address(const struct in6_addr &address, uint8_t prefix = 64, uint32_t valid_lifetime = UINT32_MAX, uint32_t preferred_lifetime = UINT32_MAX);
-	void remove_address(const struct in6_addr &address);
+	void multicast_address_was_joined(Origin origin, const struct in6_addr &address);
```
darconeous: nitpick: I normally think of joining "multicast groups", not joining "multicast addresses".

third_party/openthread/src/ncp/spinel.h
```diff
@@ -979,6 +979,9 @@ typedef enum
      */
     SPINEL_PROP_IPV6_ICMP_PING_OFFLOAD  = SPINEL_PROP_IPV6__BEGIN + 5, ///< [b]
 
+    SPINEL_PROP_IPV6_MULTICAST_ADDRESS_TABLE
```
darconeous: Wasn't there supposed to be two tables, one read-only with all multicast addresses, and one that is read/write for application-added multicast addresses? Or am I misremembering an earlier discussion?

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: Yes. I remember wondering and checking this some time ago.... :)  Any allocated/created object stays valid till end of statement.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -654,25 +644,30 @@ SpinelNCPInstance::update_node_type(NodeType new_node_type)
 	}
 }
 
-
 void
 SpinelNCPInstance::update_link_local_address(struct in6_addr *addr)
```
abtink: I agree and I was considering this, but  then decided against it for now... We have the link-local and mesh-local addresses as part of "wpanctl status" which I think is very helpful...

Within OpenThread we sort of treat the mesh-local and the thread link-local address (the one NCP calculates) as special case (e.g., user can add more link-local addresses but the NCP generated one is sort of kept/managed differently).

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -654,25 +644,30 @@ SpinelNCPInstance::update_node_type(NodeType new_node_type)
...
 	}
+
+	// Link-local address is always re-added to ensure that it is present after
+	// a reset where the address table maintained by `NCPInstanceBase` is wiped
+	// but the `mNCPLinkLocalAddress` remains valid. Note that we don't want to
+	// clear `mNcpLinkLocalAddress` so not to emit a property changed signal for
+	// `kWPANTUNDProperty_IPv6LinkLocalAddress` for an NCP reset.
+
+	if (IN6_IS_ADDR_LINKLOCAL(&mNCPLinkLocalAddress)) {
+		unicast_address_was_added(kOriginThreadNCP, mNCPLinkLocalAddress);
```
abtink: With new model, all addresses are being tracked by `NCPInstanceBase` and  if we already had the address in our table, no action will be taken. If it's a new one then we ensure to sync/push it on other interface (i.e., if NCP added it we add it on host interface and vice versa)... So technically the same address would/should not be added multiple times on the primary interface...

From plug-in sub-class we always notify our base class that an addess was added/rmoved on NCP. The base class ensures to keep track of the state in its table.

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: We can consider it. Currently OT does not filter and pass up all addresses. There may be value for debugging to see all addresses (including RLOC ones) on wpantund.
We were filtering the RLOC based addresses when the `IPV6_ADDRESS_TABLE` was parsed, I added a method to make it more explicit.

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: I agree... Will change them to `LOG_INFO`. I think they were originally `LOG_NOTICE` that's why I kept them...

src/util/TunnelIPv6Interface.cpp
abtink: Agree... will change them....

src/util/TunnelIPv6Interface.cpp
abtink: Even though looks weird, this is intentional.

On `TunnelIPv6Interface` we keep track of addresses that we are told to add in a list. If the interface is down we don't add the addresses immediately and wait and add them after the interface is brought up. For multi-cast addresses after we add them we can clear our table, for uni-cast addresses we wait for the net-link signal confirming that the address was added before removing it from our list.

(based on some  checks, on linux it seems ok to add/remove addresses while interface is down, but it does not work on mac OS).

src/wpantund/NCPInstanceBase-Addresses.cpp
abtink: I agree. I kept it as it was before for now... Thinking about this it may be good to remove it within the change.

src/wpantund/NCPInstanceBase-Addresses.cpp
abtink: The method `should_filter_address` basically let's sub-classes to filter certain addresses. I added a simple implementation in base class to avoid making it an abstract virtual function (if plugin sub-class want to avoid this). 

Spinel plugin uses this method to filter RLOC based addresses. We can consider having the addresses filtered by NCP or by the sub-class directly but I felt having the full list of NCP added addresses and showing which ones were filtered can be helpful for debugging.

src/wpantund/NCPInstanceBase.cpp
abtink: Good catch...Thanks... I have no idea how this was added here...will remove it...

src/wpantund/NCPInstanceBase.cpp
abtink: Good point! Will remove it.

src/wpantund/NCPInstanceBase.cpp
abtink: The condition here checks if old state was commissioned and new state is not, then we go ahead and issue `reset_interface()` which then erases all addresses/prefixes/etc. The main use-case is when for example, we "leave", to clear all the table and reset everything.

The new extra `ncp_state_is_joining()` is to avoid the following case:
- A already commissioned NCP is reset
- We initialize it and we end up in "COMMISIONED" state
- Then we resume, this causes a transition to "ASSOCIATING" state (for which `ncp_state_is_commissioned()` correctly returns `false`)
- This causes the addresses/routes/etc to be wiped.

third_party/openthread/src/ncp/spinel.h
abtink: You remember it correctly. I considered this as well, but realized that this requires some bigger change in how OT stores/manages the addresses.

abtink:
In the new pushed commit made some changes to address comments: 
- Removed the timeouts for unicast address entris
- Change log-level for tunnel address add/remove to `LOG_INFO`
- Remove extra code in parsing of new property.

src/ncp-spinel/SpinelNCPInstance.cpp
darconeous: 👍

src/ncp-spinel/SpinelNCPInstance.cpp
darconeous: Keeping track of them independently is certainly useful, but using them to keep the network address list in sync seems redundant.

src/ncp-spinel/SpinelNCPInstance.cpp
darconeous: The full RLOC IPv6 addresses are a thread implementation detail. Other types of networks wouldn't have them, and including them here requires a special case for thread networks. I'd rather that special case be in the NCP at some point in the near future.

src/ncp-spinel/SpinelNCPInstance.cpp
darconeous: 👍

src/util/TunnelIPv6Interface.cpp
darconeous: So this is more like a queue of pending multicast addresses to add than the list of multicast addresses...?

src/wpantund/NCPInstanceBase-Addresses.cpp
darconeous: Pinging @jwhui. We should figure out a well-defined long-term solution for how to handle this.

src/wpantund/NCPInstanceBase-Addresses.cpp
darconeous: The more I think about it the more strongly I think we should not be filtering addresses and that the NCP should only be returning to us usable IPv6 addresses in the IPv6 address list. If we want to add another property for the internal IPv6 address list then we should do that instead of adding them all to that list. Filtering out addresses is a hack.

src/wpantund/NCPInstanceBase.cpp
darconeous: 👍

third_party/openthread/src/ncp/spinel.h
darconeous: Hmm... I really need to devote a day or two to Spinel stuff this week to work out these details. I'm afraid we are coding ourselves into a corner with stuff like this.

src/util/TunnelIPv6Interface.cpp
abtink: Yes. Exactly.

third_party/openthread/src/ncp/spinel.h
abtink: I think realizing this probably require changes beyond the ncp/spinel and most likely into OT core.

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: I see your point. Let me see if we can change this.

src/wpantund/NCPInstanceBase-Addresses.cpp
abtink: I agree that NCP filtering the address would be a better solution. 

For now, we need to filter somewhere in wpantun for now since NCP does not. **I will move the filtering back in the `SpinelNCPInstance` (when we parse the IPv6_ADDRESS_TABLE) and remove the API from  the `NCPInstanceBase`....**

What I like about the base class tracking them is we get a nice full  list when we query "IPv6:AllAddresses" like this
````
IPv6:AllAddresses = [
	"fd9a:7416:beb7::ff:fe00:9800             prefix_len:64   origin:ncp      filtered:yes  valid:forever   preferred:forever"
	"fd9a:7416:beb7::ff:fe00:fc00             prefix_len:128  origin:ncp      filtered:yes  valid:forever   preferred:forever"
	"fd9a:7416:beb7:0:817e:4962:b26d:15ce     prefix_len:64   origin:ncp      filtered:no   valid:forever   preferred:forever"
	"fe80::4496:6a31:288f:c1be                prefix_len:64   origin:ncp      filtered:no   valid:forever   preferred:forever"
]
````
But I agree the better way is have two seperate spinel property (private thread-only addresses vs public addresses) which can be asscoaited with  different wpan properties or be combined into one.

**Comment by abtink** [comment]
New puhsed commit contains the following changes to this PR:
- Removed all address filtering code/methods from `NCPInstanceBase` 
- Removed adding/removing of LL and ML addresses from `SpinelNCPInstance` update_<...> method (now all addresses are  synced from same path in the code where `IPV6_ADDRESS_TABLE` is parsed).
- RLOC address filter happens is plugin `SpinelNCPInstance` class .

File: src/ncp-spinel/SpinelNCPInstance.cpp (hunk 13 above)
Comment (abtink) [inline src/ncp-spinel/SpinelNCPInstance.cpp]: Changed this in new pushed commit

File: src/wpantund/NCPInstanceBase-Addresses.cpp [hunk 15]
```diff
@@ -30,122 +30,458 @@
...
+		for (iter = mUnicastAddresses.begin(); iter != mUnicastAddresses.end(); iter++) {
+			if (!iter->second.is_from_ncp()) {
+				continue;
+			}
+
+			syslog(LOG_INFO, "UnicastAddresses: Removing %s", iter->second.get_description(iter->first).c_str());
+			mPrimaryInterface->remove_address(&iter->first, iter->second.get_prefix_len());
+
+			mUnicastAddresses.erase(iter);
+			did_remove = true;
```
darconeous: This looks like it will only remove a single address...? Was this intentional?

darconeous: Oops nevermind, I had a mental boolean logic failure.

PR #215, opened 2017-08-21
Lower the (ncpbound) spinal command logs to LOG_DEBUG

abtink:
Note that the corresponding logs for response from NCP is still
kept at LOG_INFO level.


abtink:
Closing this.

PR #221, opened 2017-08-31
Update SLAAC address for on-mesh-prefix entries independent of stable flag

abtink:


abtink:
Reviewed internally.... merging the commit.

Issue #222, opened 2017-09-11
Implement more comprehensive fuzzing


darconeous:
We need to build upon the work started in #191 to implement fuzzing of the NCP datapath.

darconeous:
@bukepo: Are you still up for this task?

darconeous:
FYI: See the [Fuzzing page](https://github.com/openthread/wpantund/wiki/Fuzzing) on the wiki for instructions on how to fuzz.

darconeous:
All done

PR #224, opened 2017-09-12
fuzzing: Initial NCP traffic fuzzing

darconeous:
This change introduces the first version of NCP traffic fuzzing, which is needed for issue #222.

Note that we still need to add an appropriate corpus.


darconeous:
I've got this working well with [my OSS-Fuzz changes](https://github.com/darconeous/oss-fuzz/tree/wpantund-fuzz/projects/wpantund) on my local machine.

@abtink or @bukepo, could you have a look at these changes so I can turn on OSS-Fuzz?

PR #228, opened 2017-09-15
spinel: Updates from OpenThread

darconeous:
This commit includes the changes to the Spinel parser from OpenThread commit 908d7dae, and includes the changes from [PR#2200][2200].

[2200]: https://github.com/openthread/openthread/pull/2200


darconeous:
I've done a lot of testing with this change and it seems pretty stable. Since it is creating a lot of issues on OSS-Fuzz, I'm going to go ahead and pull the trigger.

PR #229, opened 2017-09-18
string-utils: Fix for one-byte buffer overflow in parse_string_into_data

darconeous:
This change addresses a one-byte buffer overflow in the function which parses a hex string into data, `parse_string_into_data()`. The buffer overflow would occur when the input string consisted of a single valid hexadecimal digit, and would result in the a byte with the valid of that digit in the upper nibble (and a zero in the lower nibble) being written to an invalid location in the stack.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3396

### DIFF
--- a/src/util/string-utils.c
+++ b/src/util/string-utils.c
@@ -81,7 +81,7 @@ parse_string_into_data(uint8_t* buffer, size_t len, const char* c_str)
 		len = 0;
 	}
 
-	while (*c_str != 0) {
+	while ((*c_str != 0) && (len > 0)) {
 		char c = tolower(*c_str++);
 		if (!(isdigit(c) || (c >= 'a' && c <= 'f'))) {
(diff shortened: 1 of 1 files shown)


src/util/string-utils.c
```diff
@@ -81,7 +81,7 @@ parse_string_into_data(uint8_t* buffer, size_t len, const char* c_str)
 		len = 0;
 	}
 
-	while (*c_str != 0) {
+	while ((*c_str != 0) && (len > 0)) {
```
bukepo: Do we need to verify `c_str` is null-terminated and even-length?

darconeous: Yes. `len` is for the size of `buffer`. `c_str` is a zero-terminated c-string, and has no explicit length. We must check both at each step.

PR #232, opened 2017-09-18
wpantund-fuzz: Ignore SocketError exceptions.

darconeous:
Throwing SocketError exceptions is an acceptable error condition.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3392


bukepo:
@darconeous I don't have the permission to see the link in the post.

darconeous:
You don't need to be able to view it to review. It will become public at some point.

darconeous:
Going ahead and merging so that I can get this issue included in the next fuzzer build.

PR #233, opened 2017-09-19
ncp-spinel: Fix for NULL dereference on malformed packet

darconeous:
Found via [oss-fuzz][] (issue [3394][chromium]).

[chromium]: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3394
[oss-fuzz]: https://oss-fuzz.com/v2/testcase-detail/6006849620934656

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -2396,18 +2396,20 @@ SpinelNCPInstance::handle_ncp_spinel_value_inserted(spinel_prop_key_t key, const
 	} else if (key == SPINEL_PROP_THREAD_ON_MESH_NETS) {
 		struct in6_addr *prefix = NULL;
 		uint8_t prefix_len = 0;
-		bool stable;
+		bool stable = false;
 		uint8_t flags = 0;
-		bool is_local;
+		bool is_local = false;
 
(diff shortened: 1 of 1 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2396,18 +2396,20 @@ SpinelNCPInstance::handle_ncp_spinel_value_inserted(spinel_prop_key_t key, const
 	} else if (key == SPINEL_PROP_THREAD_ON_MESH_NETS) {
 		struct in6_addr *prefix = NULL;
 		uint8_t prefix_len = 0;
-		bool stable;
+		bool stable = false;
 		uint8_t flags = 0;
-		bool is_local;
+		bool is_local = false;
 
 		spinel_datatype_unpack(value_data_ptr, value_data_len, "6CbCb",
```
bukepo: I'm not sure if we should also check the return value here, which ensures the value of `prefix_length`, `stable`, `flags`, and `is_local` is properly parsed.

darconeous: You bring up a good point. I can't remember which fields we consider to be optional for this property. In this case we are considering all but the first to be "optional".

This is likely a wider subject that needs some more careful thought, since there are a lot of other cases that this could apply to. We can track that with a separate issue (Will file shortly).

PR #235, opened 2017-09-21
Add option: SLAAC for auto added prefix.

prfi:
This pull request adds option SLAAC for auto added prefixes. This option can be used to setup IPv6 addresses for prefixes delegated via DHCPv6-PD on the BR.

### DIFF
--- a/src/wpantund/wpan-properties.h
+++ b/src/wpantund/wpan-properties.h
@@ -82,6 +82,7 @@
 #define kWPANTUNDProperty_IPv6MeshLocalPrefix                   "IPv6:MeshLocalPrefix"
 #define kWPANTUNDProperty_IPv6AllAddresses                      "IPv6:AllAddresses"
 #define kWPANTUNDProperty_IPv6MulticastAddresses                "IPv6:MulticastAddresses"
+#define kWPANTUNDProperty_IPv6SetSLAACForAutoAddedPrefix        "IPv6:SetSLAACForAutoAddedPrefix"
 
 #define kWPANTUNDProperty_ThreadRLOC16                          "Thread:RLOC16"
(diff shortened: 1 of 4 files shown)


darconeous:
Mind rebasing and adding a more detailed explanation in the commit log message? Specifically, what exactly the new property does and why it is needed.

Also, why use the `Daemon:` property category? Those properties are specific to the general non-networking related behavior of `wpantund`. Seems like `Thread:`, `OpenThread:` or even `IPv6:` would be more appropriate.

I'd also like to get @abtink's input before pulling the trigger, since my focus recently has been on Android.

src/wpantund/wpan-properties.h
```diff
@@ -48,6 +48,7 @@
 #define kWPANTUNDProperty_DaemonAutoDeepSleep                   "Daemon:AutoDeepSleep"
 #define kWPANTUNDProperty_DaemonFaultReason                     "Daemon:FaultReason"
 #define kWPANTUNDProperty_DaemonSetDefRouteForAutoAddedPrefix   "Daemon:SetDefaultRouteForAutoAddedPrefix"
+#define kWPANTUNDProperty_DaemonSetSLAACForAutoAddedPrefix      "Daemon:SetSLAACForAutoAddedPrefix"
```
darconeous: It feels like we should use a different property category instead of `Daemon:`. How about `IPv6:`?

I realize @abtink set a precedent with line 50 (from #209), but I'm not sure it is the right one. @abtink, feel free to chime in with your thoughts.

abtink: I am fine with changing them both to `"IPv6:<name>"`

prfi:
I've rebased commit and I've extended commit log message.

darconeous:
Manually merged via c3f43e8.

PR #239, opened 2017-09-22
npc-spinel: Allow user to control filtering of RLOC addresses

abtink:
By default, the RLOC IPv6 addresses from NCP are filtered by wpantund
and are not added on the primary wpan interface. This commits adds a
new wpan property "Thread:Config:FilterRLOCAddresses" to allow user
to control/change this behavior. It is expected that this configuration
property  is set as part of `wpantund.conf` settings.


Nopik:
Any idea why this PR failed on Travis? `dbus-1: package not found` seems like broken testsuite, not a code problem..

abtink:
Seems like master is also seeing the same [error](https://travis-ci.org/openthread/wpantund/builds/282951718).

darconeous:
That's pretty weird. Maybe something is busted with pkg-config on homebrew? I can make those builds optional for now.

darconeous:
Rebase and travis should be nice to you.

PR #242, opened 2017-10-01
Change PSK type in joiner_add API to byte array

GiedriusM:
This is accompanying change for https://github.com/openthread/openthread/pull/2185/.

### DIFF
--- a/src/wpanctl/tool-cmd-commissioner.c
+++ b/src/wpanctl/tool-cmd-commissioner.c
@@ -61,6 +61,8 @@ int tool_cmd_commissioner(int argc, char* argv[])
 	const char* ext_addr = NULL;
 	const char* psk = NULL;
 	int psk_len = 0;
+	uint8_t pskd[20];
+	uint32_t pskd_length = sizeof(pskd);
 	uint32_t joiner_timeout = DEFAULT_JOINER_TIMEOUT;
 	dbus_bool_t enabled = false;
 	const char* invalid_psk_characters = INVALID_PSK_CHARACTERS;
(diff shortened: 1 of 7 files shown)


src/wpanctl/tool-cmd-commissioner.c
```diff
@@ -323,6 +325,11 @@ int tool_cmd_commissioner(int argc, char* argv[])
 				goto bail;
 			}
 
+			// XXX: interpret PSK string as binary PSKd
+			memcpy(pskd, psk, psk_len);
```
abtink: Possible memory corruption...`pskd` is defined as `uint8_t pskd[20];` (20 bytes long)... what is psk_len is larger than 20 bytes... ? 
We check for it to be less than `PSK_MAX_LENGTH` which is 32...

src/wpanctl/tool-cmd-commissioner.c
```diff
@@ -61,6 +61,8 @@ int tool_cmd_commissioner(int argc, char* argv[])
 	const char* ext_addr = NULL;
 	const char* psk = NULL;
 	int psk_len = 0;
+	uint8_t pskd[20];
```
abtink: Magic number here... (why not use `PSK_MAX_LENGTH`)....

darconeous:
See https://github.com/openthread/openthread/pull/2185#issuecomment-336235533

GiedriusM:
Just to clarify. I've set pskd to be PSK_MAX_LENGTH, so it would solve both issues. However, in theory pskd should be 20 bytes and it should use the OT_PSKD_MAX_SIZE definition from openthread and use base32-thread decode function to convert from PSK (joiner credential) into PSKd. The fact that there's no precise specification for base32-thread encoding and that openthread test harness uses and passes PSKd directly (not that much related to wpantund), the decision was to fix core layers to pass binary PSKd, but just copy the PSK (Joiner credential) as an incremental change and fix (implement) the missing base32-thread decoding at the user input level once other issues are resolved.

More details and discussions can be found https://github.com/openthread/openthread/issues/2156 and https://github.com/openthread/openthread/pull/2185.

jwhui:
This is being tracked in [SPEC-742](https://threadgroup.atlassian.net/browse/SPEC-742).  In my discussions with the Thread Group, there seems to be general agreement that the Joiner Device Credential and PSKd are the same in both information and encoding.  There does not seem to be any use case for the binary version.  As a result, I'll close this PR.

Issue #245, opened 2017-10-16
Travis macOS builds are failing.


darconeous:
First noticed on #239, it appears that all macOS Travis builds are failing:

```
checking pkg-config is at least version 0.9.0... yes
checking for DBUS... no
configure: error: Package requirements (dbus-1 >= 1.4) were not met:
No package 'dbus-1' found
Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
Alternatively, you may set the environment variables DBUS_CFLAGS
and DBUS_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
+die
+echo ' *** ERROR: '
 *** ERROR: 
+exit 1
```

darconeous:
Will go ahead and make macOS failures be ignored for now.

darconeous:
Moving to P2 now that this is unblocking other commits.

darconeous:
It seems like MacOS travis builds are magically working again.

PR #246, opened 2017-10-16
Temporary work-around for #245: Ignore macOS build failures

darconeous:
This is a temporary work-around for issue #245.


darconeous:
Non-skipped builds are looking good, going ahead and merging.

PR #247, opened 2017-10-19
Handle and log child entry inserted/removed spinel messages from NCP

abtink:
This commit contains the following changes: (a) It updates the
`SpinelNCPTaskGetNetworkTopology` to provide a new  static method for
parsing of a single child table entry (which is also used in the
method for parsing the entire child table). (b) It also updates the
`handle_ncp_spinel_value_inserted/removed()` methods to parse and log
`CHILD_TABLE` updates from NCP.


abtink:
The related PR in OT: https://github.com/openthread/openthread/pull/2280

Issue #244, opened 2017-10-14
WPanTunD doesnt support multicast on OSX


Nopik:
When running `wpantund` on OSX, the following error shows:

```
Oct 13 09:29:46  wpantund[68138] <Warning>: ../util/netif-mgmt.c:479: Requirement Failed (Undefined error: 0)
Oct 13 09:29:46  wpantund[68138] <Warning>: Failed to join multicast group "ff03::1"
```

and as a result the `wpantund` doesn't support multicasts.

Maybe it is related to the lack of `SIOGIFINDEX` support?

PR #248, opened 2017-10-25
Multicasts fix for OSX

mwilniewiec:
Fix for #244

SIOGIFINDEX doesn't exist on OSX and no interface index was retrieved.
This was the under the hood problem of join_multicast_group() failure.


mwilniewiec:
I signed it!

mwilniewiec:
Im using osx so i made a safe fix. True it seems that if_nametoindex is available on all platforms but it would be nice to verify that first.

PR #251, opened 2017-11-09
Support for getting/changing Thread Operational Dataset

abtink:
This commit adds support in wpantund and wpanctl to get, set and
modify the Active/Pending Thread Operational Dataset on NCP. A new
class `ThreadDataset` is defined to represent all the fields within a
Thread Operational Dataset.

New wpan properties are defined to allow access to all dataset fields
(in a human-readable string format or as value-map dictionary). A new
set of wpan properties are added to allow user to change different
fields within a local (to wpantund) Dataset container. A set of new
(wpanctl) commands can then be used to get/push the local Dataset on
NCP as either Active or Pending Dataset.

Related PR in OT: https://github.com/openthread/openthread/pull/2327


abtink:
PR reviewed and verified internally.

PR #256, opened 2017-12-06
Remove extra use of lock property `ALLOW_LOCAL_NET_DATA_CHANGE`

abtink:
The lock property `SPINEL_PROP_THREAD_ALLOW_LOCAL_NET_DATA_CHANGE` should be used when the underlying spinel propery would change the Thread Network Data (e.g., on-mesh prefixes or off-mesh routes). This commit fixes the extra/incorrect use of this lock property for other properties.

This should address https://github.com/openthread/openthread/issues/2110.


abtink:
Reviewed internally.

PR #258, opened 2017-12-06
[ncp-spinel] Transforming spinel frames between AP and NCP

kbogucki:
NCP Spinel Transformer allows to transform spinel frames sent between
Application Processor (AP) and Network Co-Processor (NCP).

Spinel frames can be transformed by using transformer library
(or libraries) implementing spinel_transformer.hpp. Libraries
can be specified using option --with-ncp-spinel-transformer-libs.
Addidionaly, transformer's outbound buffer can be changed if needed,
using option --with-ncp-spinel-transformer-outbound-buffer-size.


kbogucki:
I signed it!

kbogucki:
Signed again!

kbogucki:
Related PR in openthread repo: https://github.com/openthread/openthread/pull/2402

kbogucki:
Spinel Transformer example implementation (XOR):
https://gist.github.com/kbogucki/d16d2336bdb00efd196e37fe30b7e597

Nopik:
@jwhui the implementation seems to be finished, wpantund side of it. Could you take a look on it, please?

src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
```diff
@@ -299,6 +303,16 @@ SpinelNCPInstance::ncp_to_driver_pump()
 			goto on_error;
 		}
 
+#if OPENTHREAD_ENABLE_NCP_SPINEL_ENCRYPTER
+		size_t inboundFrameTransformedLen = sizeof(mInboundFrame);
+		if (!SpinelEncrypter::DecryptInbound(mInboundFrame, mInboundFrameSize, mInboundFrame, &inboundFrameTransformedLen))
```
abtink: We are passing the same `mInBoundFrame` buffer as both input and output buffer to `DecryptInbound()`  assuming its implementation can handle this correctly. If that's the case, how about if we change the `DecryptInbound` to only accept one buffer pointer and does in-place description. This way it forces the implementer to be careful about this.

I am fine with the `DecryptInbound()` to expect an input and output buffer. In that case, I would suggest we send it two different buffers. How about if we have something like this (use a local buffer to store transformed frame - since on wpantund we are not as resource-limited)....
```
#if OPENTHREAD_ENABLE_NCP_SPINEL_ENCRYPTER
        size_t inboundFrameTransformedLen = sizeof(mInboundFrame);
        uint8_t inboundFrameTransforemed[SPINEL_FRAME_BUFFER_SIZE];
        if (!SpinelEncrypter::DecryptInbound(mInboundFrame, mInboundFrameSize, inboundFrameTrasnformed, &inboundFrameTransformedLen))
        {
        ....
        }
        memcpy(mInboundFrame, inboundFrameTransformed, inboundFrameTrasnformedLen);
        mInboundFrameSize = inboundFrameTrasnformedLen;
```

src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
```diff
@@ -505,6 +519,17 @@ SpinelNCPInstance::driver_to_ncp_pump()
 			spinel_ssize_t i;
 			uint8_t byte;
 			uint16_t crc(0xFFFF);
+
+#if OPENTHREAD_ENABLE_NCP_SPINEL_ENCRYPTER
+			size_t outboundFrameTransformedLen = sizeof(mOutboundBuffer);
+			if (!SpinelEncrypter::EncryptOutbound(mOutboundBuffer, mOutboundBufferLen, mOutboundBuffer, &outboundFrameTransformedLen))
```
abtink: Same suggestion as above for the `EncryptOutbound()`.

abtink:
Overall, looks good. I have one suggestion about the calling of Encrypt/Decrypt with separate buffers

src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
abtink: Sounds good. Thanks.

kbogucki: Done.

PR #261, opened 2017-12-21
Adding new property "Thread:ChildTable:Addresses"

abtink:
This commit adds a new wpan property "Thread:ChildTable:Addresses"
which is mapped to SPINEL_PROP_THREAD_CHILD_TABLE_ADDRESSES. This
property provides the list of all addresses associated with every
child including any registered IPv6 addresses.

This commit also simplifies class SpinelNCPTaskGetNetworkTopology
implementation by adding a new `parse_table()` method which includes
the common code used for parsing different spinel properties related
to network topology (child table, neighbor table, router table, etc).

Related PR in OT: https://github.com/openthread/openthread/pull/2434


abtink:
Reviewed and verified internally

PR #265, opened 2018-01-08
Fix problem with building wpantund.

prfi:
This pull request solves problem with building latest wpantund on the RPi3 platform. During building the border router I've got an error which says that cannot bind bitfields variables with their names. e.g. ```SpinelNCPTaskGetNetworkTopology.cpp:698:83: error: cannot bind bitfield ((nl::wpantund::SpinelNCPtaskGetNetworkTopology::TableEntry*)this)->nl::wpantund::SpinelNCPTaskGetNetworkTopology::TableEntry::mRxOnWhenIdle to bool&```.

### DIFF
--- a/src/ncp-spinel/SpinelNCPTaskGetNetworkTopology.cpp
+++ b/src/ncp-spinel/SpinelNCPTaskGetNetworkTopology.cpp
@@ -669,7 +669,7 @@ bail:
 }
 
 ValueMap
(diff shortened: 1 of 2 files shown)


abtink:
Thanks for the fix. We noticed this as well on certain platform. It's interesting that making the method `const` addresses the issues. Curios why on some platform it tries to to bind it as `bool &`.

PR #266, opened 2018-01-08
Additional diagnostic updates for oss-fuzz-3565

darconeous:
[oss-fuzz-3565](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=3565) describes an unreproducable (but frequently occuring) global buffer overflow discovered by fuzzing. The immediate cause of the overflow is the failure of `fd_set_merge()` to respect the size boundaries of an `fd_set`, which is limited to accounting to no more than `FD_SETSIZE` file descriptors. That issue was fixed by #263. However, the fact that we are getting anywhere close to the `FD_SETSIZE` file descriptor limit indicates that we may very well be leaking file descriptors.

Pull request #263 added some checks to detect file descriptor leaks, but this pull request introduces better checking that should be able to ferret out the exact sorts of test cases that induce leaking file descriptors.

### DIFF
--- a/src/wpantund/wpantund-fuzz.cpp
+++ b/src/wpantund/wpantund-fuzz.cpp
@@ -52,26 +52,14 @@ count_open_fds()
 
 		if (fd >= 0) {
 			count++;
-			close(fd);
+			if (close(fd) < 0) {
+				perror("count_open_fds");
+				fuzz_trap();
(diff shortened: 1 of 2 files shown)


darconeous:
Confirmed with oss-fuzz on local machine.

PR #272, opened 2018-01-29
Add new wpan property "NCP:CCAFailureRate"

abtink:
The wpan property is mapped to newly added NCP spinel property
`SPINEL_PROP_MAC_CCA_FAILURE_RATE` which provides the CCA
(Clear Channel Assessment) failure rate. The value is provided
as a `uint16_t` with 0 mapping to 0% error rate, and 0xffff
indicating 100% failure rate.


abtink:
This change was verified/tested internally.

PR #274, opened 2018-02-03
Get full IPv6 address table on NCP state change to "associated"

abtink:
This commit adds `SPINEL_PROP_IPV6_ADDRESS_TABLE` to the list of
properties that are fetched from `handle_ncp_state_change()` after
an NCP state change to "associated".


abtink:
Verified internally.

PR #276, opened 2018-02-16
Properties to control how off-mesh-routes are managed on primary interface

abtink:
This commit adds two new `wpantund` daemon configuration properties:

- `"Daemon:OffMeshRoute:AutoAddOnInterface"` - This boolean property determines whether `wpantund` should manage the routes on the primary interface. When set to `true` wpantund will add/remove off-mesh routes provided by devices within the network on the host interface. Otherwise, it is expected the routes would be managed externally. By default this feature is enabled (`true`).

- `"Daemon:OffMeshRoute:FilterSelfAutoAdded"` - This boolean property controls how the off-mesh-routes are managed on the primary interface (this is applicable only if the `AutoAddOnInterface` feature is enabled). 
This property impacts the behavior where the same off-mesh route is provided by multiple devices within the network including the device itself. When set to `true`, self-added off-mesh routes are always filtered  and never added on the interface (independent of the priority levels at which they are added by device or others). If it is set to `false`,  then priority of routes are considered and the off-mesh-route is  added on the interface if another device within the network provides the same route at a higher preference level than the self added one.  By default this is enabled (`true`).

### DIFF
--- a/src/wpantund/NCPInstanceBase.h
+++ b/src/wpantund/NCPInstanceBase.h
@@ -479,6 +479,33 @@ protected:
 	bool mSetDefaultRouteForAutoAddedPrefix;
 	bool mSetSLAACForAutoAddedPrefix;
 
+	// This boolean flag determines whether wpantund should manage the routes
+	// on the primary interface. When set to `true` wpantund will add/remove
+	// off-mesh routes provided by devices within the network on the host
+	// interface. By default it is enabled (`true`).
+	//
+	bool mAutoAddOffMeshRoutesOnInterface;
+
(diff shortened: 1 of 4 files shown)


src/wpantund/NCPInstanceBase.h
```diff
@@ -479,6 +479,33 @@ class NCPInstanceBase : public NCPInstance, public EventHandler {
 	bool mSetDefaultRouteForAutoAddedPrefix;
 	bool mSetSLAACForAutoAddedPrefix;
 
+	// This boolean flag determines whether wpantund should manages the routes
```
smides-nest: nit: s/manages/manage

PR #278, opened 2018-02-23
Add config option "Deamon::IPv6::AutoUpdateInterfaceAddrsOnNCP"

abtink:
This commit adds a new configuration option to determine whether
wpantund should listen for unicast IPv6 address-added/removed events
from the primary interface and then update the addresses on the NCP.
By default this feature is enabled. It can be changed using a
corresponding configuration wpantund property.


abtink:
Verified internally.

PR #279, opened 2018-02-27
Add config option "Daemon:IPv6:FilterUserAddedLinkLocal"

abtink:
This commit adds a new configuration option to determine whether
wpantund should skip adding user (or interface) originated link-local
IPv6 addresses on NCP. By default this feature is enabled. The
behavior can  be changed using wpantund configuration property
"Daemon:IPv6:FilterUserAddedLinkLocal".


abtink:
Verified internally.

PR #280, opened 2018-02-28
Fix for usage of uninitialized memory

darconeous:
This change addresses a use-of-uninitialized-memory issue discovered by oss-fuzz.

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=4616


darconeous:
Going ahead and merging this for now so that the fuzzer can start having a look at it.

PR #282, opened 2018-03-13
add mesh local prefix setting support

librasungirl:
Add support to explicitly configure mesh local prefix.

I am not very clear why configuring the network parameters again when forming the network, but, one issue when adding `IPv6:MeshLocalPrefix` set operation is that it would be replaced by the one coming from XPANID because mesh local prefix would be configured again when configuring XPANID again during forming (see https://github.com/openthread/openthread/blob/master/src/core/api/thread_api.cpp#L83). So another change in this PR is to prefer the mesh local prefix explicitly configured by property to the one from mesh local address.

@abtink, please help to evaluate if it has any possible issue I didn't notice. Thanks.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -2039,6 +2039,30 @@ SpinelNCPInstance::property_set_value(
 				);
 			}
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_IPv6MeshLocalPrefix)) {
+			struct in6_addr address = any_to_ipv6(value);
+			char address_str[INET6_ADDRSTRLEN] = "::";
+
+			inet_ntop(AF_INET6, (const void *)&address, address_str, sizeof(address_str));
+
+			fprintf(stderr, "Using prefix \"%s\"\n", address_str);
+
+			Data command =
+				SpinelPackData(
+					SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_DATA_S),
+					SPINEL_PROP_IPV6_ML_PREFIX,
+					address.s6_addr,
+					8
+				);
+
+			mSettings[kWPANTUNDProperty_IPv6MeshLocalPrefix] = SettingsEntry(command);
+
+			start_new_task(SpinelNCPTaskSendCommand::Factory(this)
+				.set_callback(cb)
+				.add_command(command)
+				.finish()
+			);
+
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_ThreadCommissionerEnabled)) {
 			bool isEnabled = any_to_bool(value);
 
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -2040,14 +2040,14 @@ SpinelNCPInstance::property_set_value(
 			}
 
 		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_IPv6MeshLocalPrefix)) {
-			struct in6_addr address = any_to_ipv6(value);
+			struct in6_addr addr = any_to_ipv6(value);
 
 			Data command =
 				SpinelPackData(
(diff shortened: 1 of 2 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2039,6 +2039,30 @@ SpinelNCPInstance::property_set_value(
 				);
 			}
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_IPv6MeshLocalPrefix)) {
+			struct in6_addr address = any_to_ipv6(value);
+			char address_str[INET6_ADDRSTRLEN] = "::";
+
+			inet_ntop(AF_INET6, (const void *)&address, address_str, sizeof(address_str));
+
+			fprintf(stderr, "Using prefix \"%s\"\n", address_str);
```
abtink: How about removing the log here?
The mesh-local prefix change will be logged (using `syslog`) on the spinel `VALUE_IS` reply for the `VALUE_SET` command. This happens from `SpinelNCPInstance::handle_ncp_spinel_value_is()` call here: https://github.com/openthread/wpantund/blob/master/src/ncp-spinel/SpinelNCPInstance.cpp#L2599

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2039,6 +2039,30 @@ SpinelNCPInstance::property_set_value(
...
+
+			Data command =
+				SpinelPackData(
+					SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_DATA_S),
+					SPINEL_PROP_IPV6_ML_PREFIX,
+					address.s6_addr,
+					8
+				);
+
+			mSettings[kWPANTUNDProperty_IPv6MeshLocalPrefix] = SettingsEntry(command);
```
abtink: I am not sure if want to save the Mesh-Local Prefix in the `mSettings`?
This is not done for channel, PANId, XPANID, NetworkName, ...and other similar network info.

NCP/OT already saves the mesh-local prefix as part of Operational Dataset (in flash in its settings).  The `mSettings` in `wpantund` is intended for configurations settings (e.g., jamming detection settings) that we need to persist over NCP resets but the are not saved by the OT/NCP itself.

src/ncp-spinel/SpinelNCPInstance.cpp
librasungirl: Make sense. Updated.

src/ncp-spinel/SpinelNCPInstance.cpp
librasungirl: Updated. Thanks for the explanations!

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -2039,6 +2039,23 @@ SpinelNCPInstance::property_set_value(
...
 
+		} else if (strcaseequal(key.c_str(), kWPANTUNDProperty_IPv6MeshLocalPrefix)) {
+			struct in6_addr address = any_to_ipv6(value);
+
+			Data command =
+				SpinelPackData(
+					SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_DATA_S),
+					SPINEL_PROP_IPV6_ML_PREFIX,
+					address.s6_addr,
+					8
```
abtink: The code here is encoding the property as a data blob of 8 bytes. But `spinel.h` and docs here [spinel-prop-ipv6.md](https://github.com/openthread/openthread/blob/master/doc/spinel-protocol-src/spinel-prop-ipv6.md#prop-98-prop_ipv6_ml_prefix-prop-ipv6-ml-prefix), define `SPINEL_PROP_IPV6_ML_PREFIX` with encoding of `[6C]` (an IPv6 address followed by a prefix length). Let's change this to match. Here is example of this from OpenThread' `NcpBase` [set handler](https://github.com/openthread/openthread/blob/master/src/ncp/ncp_base_mtd.cpp#L1011).

librasungirl: Thanks. Updated.

PR #289, opened 2018-04-11
Add support for NCP MCU power state control and deep-sleep mode

abtink:

It adds support for getting/setting a new property "NCP:MCUPowerState"
in wpantund. This property is mapped to the recently added spinel
property `SPINEL_PROP_MCU_POWER_STATE` which can be used to control
the NCP's MCU power state (capability `SPINEL_CAP_MCU_POWER_STATE`
indicates support for this property and feature on NCP).

The `SpinelNCPTaskDeepSleep` and `SpinelNCPTaskWake` implementations
are updated to adopt/use the new spinel property (only when the NCP is
indicating its capability for this feature).

The commit also adds logic to manage wpantund state changes between
`OFFLINE` and `DEEP_SLEEP` based on changes to `MCU_POWER_STATE`.

Finally, `SpinelNCPControlInterface::begin_low_power()` is changed to
be an alias for setting the property "NCP:MCUPowerState" to low-
power.

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -52,6 +52,7 @@
 #include "SpinelNCPTaskLeave.h"
 #include "SpinelNCPTaskScan.h"
 #include "SpinelNCPTaskPeek.h"
+#include "SpinelNCPTaskDeepSleep.h"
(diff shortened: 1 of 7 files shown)


abtink:
Feature was verified internally and through `tornaj` test-case (please see https://github.com/openthread/openthread/pull/2659)

PR #291, opened 2018-04-12
Fix incorrect comparision in `SpinelNCPTaskGetNetworkTopology`

abtink:
Fix incorrect use of `=` instead of `==`.


abtink:
Reviewed internally.

PR #293, opened 2018-04-14
Allow auto-resume from wpan "offline:deep-sleep" state

abtink:
This commit fixes an issue with the auto-resume logic not starting
when the NCP starts with MCU in low-power (wpan state starting as
"offline:deep-sleep". It changes the test before starting the resume
process to explicitly check the `mIsCommisionned` flag and relax
the wpan state check to include any offline states.


abtink:
This was tested and verified internally.

PR #294, opened 2018-04-20
Enable adding/removing multicast IPv6 addresses

abtink:
This commit adds new feature to allow user to add/remove multicast
IPv6 addresses by issuing insert/remove using the wpan property
"IPv6:MulticastAddresses". The multicast IPv6 address will be updated
on NCP and also on the primary network interface.


abtink:
Verified internally.

PR #299, opened 2018-05-17
Add support for new NCP log output model and spinel property

abtink:
This commit adds support for the new NCP log output model. It adds
support for the newly added spinel property `SPINEL_PROP_STREAM_LOG`
which provides streaming of formatted log string from NCP along with
an optional metadata structure. OpenThread log level and log region
are included as part of the metadata. `wpantund` parses the metadata
and appends a human-readable prefix string to the outputted log line
in syslog.

Related PR in OT: https://github.com/openthread/openthread/pull/2722


abtink:
This was verified internally.

PR #305, opened 2018-06-06
Add new wpan property "Thread:RouterSelectionJitter"

librasungirl:
This commit adds new wpan property to configure the jitter
for Router-Capable device to decide whether to become Router or
stay in REED role.


librasungirl:
@abtink Please help to review it. Thanks.

PR #309, opened 2018-06-14
Replace TMF proxy with UDP proxy

bukepo:
This PR replaces TMF proxy with UDP proxy, which helps moving border agent into OpenThread. The changes in OpenThread can be reviewed at https://github.com/openthread/openthread/pull/2771

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -520,10 +520,6 @@ SpinelNCPInstance::get_supported_property_keys()const
 		properties.insert(kWPANTUNDProperty_ChannelManagerFavoredChannelMask);
 	}
 
-	if (mCapabilities.count(SPINEL_CAP_THREAD_UDP_PROXY)) {
-		properties.insert(kWPANTUNDProperty_UdpProxyEnabled);
-	}
-
 	if (mCapabilities.count(SPINEL_CAP_NEST_LEGACY_INTERFACE))
 	{
 		properties.insert(kWPANTUNDProperty_NestLabs_LegacyMeshLocalPrefix);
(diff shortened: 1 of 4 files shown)


bukepo:
@abtink rebased.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -4062,6 +4099,45 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
...
+			// append frame
+			data.append(frame_ptr, frame_len);
+			// pack the locator in big endian.
+			data.push_back(peer_port >> 8);
+			data.push_back(peer_port & 0xff);
+			data.append(peer_addr->s6_addr, sizeof(*peer_addr));
+			// pack the port in big endian.
+			data.push_back(sock_port >> 8);
+			data.push_back(sock_port & 0xff);
+			signal_property_changed(kWPANTUNDProperty_UdpProxyStream, data);
```
abtink: How about providing a get handler for the `kWPANTUNDProperty_UdpProxyStream` property to give us the last received UDP stream (last one for which we have emitted `signal_property_changed`).
We can save the `data` from last received UDP stream here in some member variable and return it from the get-handler. This way users that only parse the property name (and not the value) from `signal_property_changed` would have a way to get the frame.

Thoughts?

bukepo: Do you have a specific use case in mind for this? Even though users only parse the property name first, they are still able to further parse the value, aren't they? And it seems we cannot guarantee the last received UDP stream is still the original one when users ask for that through another d-bus call.

abtink: I was thinking of testing as the main use-case for this. 
I agree users should parse the emitted dbus signal to get both name and value so that they ensure that they don't miss any frames.

bukepo: Thanks for the clarification. How about we implement that when we actually use it? I think it's better to keep a PR small.

abtink: Sounds good to me. :+1:

PR #315, opened 2018-06-21
Remove extra `dbus_message_ref()` in DBus APIs methods

abtink:
In some of `DBusIPCAPI_v1` or `DBusIPCAPI_v0` interface methods
`dbus_message_ref()` was called twice on the DBus `message`. This
commit removes the extra call.


bukepo:
LGTM:+1:

PR #319, opened 2018-06-26
Add autogenerated version file to git ignore

EugeneGonzalez:
src/version.c isn't in gitignore, which can cause issues for automatic
syncs. The automatic sync fail will due to uncommitted changes even though
this file is generated by the build system and should be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ doxygen.cfg
 src/wpantund/wpantund
 src/wpanctl/wpanctl
 .dirstamp
+src/version.c
 src/wpanctl/version.c
 wpantund.xcodeproj/project.xcworkspace/xcshareddata/wpan-tunnel-driver.xccheckout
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -48,6 +48,7 @@ doxygen.cfg
 src/wpantund/wpantund
 src/wpanctl/wpanctl
 .dirstamp
+src/version.c
```
abtink: Thanks for your PR. I see the file at `src/wpantund/version.c`?

EugeneGonzalez: Ah, I think this is caused by the difference between Android.mk and src/Makefile.am.
Android.mk generates src/version.c on line 75 where as src/Makefile.am generates  $(top_builddir)/$(subdir)/version.c on line 47. I can update the PR to include both locations.

abtink: Sounds good. Thanks.

PR #330, opened 2018-07-10
Add 'joiner' wpanctl command and joiner properties

librasungirl:
This commit adds new 'joiner' wpanctl command to control commissioning process;
and adds "Thread:Joiner:State" spinel property to get joiner state.


src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -328,6 +329,39 @@ SpinelNCPControlInterface::joiner_add(
...
+{
+	mNCPInstance->start_new_task(SpinelNCPTaskSendCommand::Factory(mNCPInstance)
+			.set_callback(cb)
+			.add_command(SpinelPackData(
+					SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_BOOL_S),
+					SPINEL_PROP_NET_STACK_UP,
+					true
+					))
+			.finish()
+			);
```
abtink: `joiner_attach` is setting `NET_STACK_UP` property to `true`?

Can we use the existing `SpinelNCPControlInterface::attach()` instead?

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -328,6 +329,39 @@ SpinelNCPControlInterface::joiner_add(
...
+}
+
+void
+SpinelNCPControlInterface::joiner_commission(
+		bool action,
+		const char *psk,
+		const char *provisioning_url,
+		CallbackWithStatus cb
+) {
+	syslog(LOG_NOTICE, "action %d psk %s provisioning_url %s", action, psk, provisioning_url);
```
abtink: nit: adding "JoinerCommision: action %d..." in the log?

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+	CallbackWithStatusArg1 cb,
+	bool action,
+	const char *psk,
+	const char *provision_url
+	):	SpinelNCPTask(instance, cb), mAction(action)
+{
+	if (psk)
+	{
+		strncpy(mPsk, psk, sizeof(mPsk));
+		mPsk[strlen(mPsk)] = 0;
```
abtink: how about defining `mPsk and `mProvisioningUrl` as `std::string`?

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+	if (strlen(mPsk) > 0)
+	{
+		frame.append(reinterpret_cast<const uint8_t*>(mPsk), strlen(mPsk) + 1);
+	}
+
+	if (strlen(mProvisioningUrl) > 0)
+	{
+		frame.append(reinterpret_cast<const uint8_t*>(mProvisioningUrl), strlen(mProvisioningUrl) + 1);
+	}
```
abtink: Using the `SpinelPackData()` to do the spinel packing instead?

```
// for example, when we have both psk and url:

mNextCommand = SpinelPackData(
	SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(
		SPINEL_DATATYPE_BOOL_S 
		SPINEL_DATATYPE_UTF8_S 
		SPINEL_DATATYPE_UTF8_S
	),
	SPINEL_PROP_THREAD_JOINER_COMMISSION,
	action,
	mPsk,
	mProvisioningUrl
);
```

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+	syslog(LOG_INFO, "Joiner commission %s (psk: %s) (provisioning_url %s)", mAction ? "start" : "stop", mPsk, mProvisioningUrl);
+
+	mNextCommand = SpinelPackData(
```
abtink: Thinking more about this, how about if we always include the PSK and Provisioning URL in the encoding of the spinel propeert (`bUU`)?  If there is no value, we can include an empty string instead. This makes the logic in OpenThread NCP (parsing in the set handler) simpler as well.

src/ncp-spinel/SpinelNCPControlInterface.cpp
abtink: how about removing this log, since we have similar one in `SpinelNCPTaskJoinerCommission`?

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+		strncpy(mProvisioningUrl, provision_url, sizeof(mProvisioningUrl));
+		mProvisioningUrl[strlen(mProvisioningUrl)] = 0;
+	}
+}
+
+void
+nl::wpantund::SpinelNCPTaskJoinerCommission::finish(int status, const boost::any& value)
+{
+	SpinelNCPTask::finish(status, value);
+}
```
abtink: Since we are only calling base class `finish()` here how about not overriding `finish()`?  (inherited method should work)

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+	if (strlen(mPsk) > 0)
+	{
+		frame.append(reinterpret_cast<const uint8_t*>(mPsk), strlen(mPsk) + 1);
+	}
+
+	if (strlen(mProvisioningUrl) > 0)
+	{
+		frame.append(reinterpret_cast<const uint8_t*>(mProvisioningUrl), strlen(mProvisioningUrl) + 1);
+	}
+	syslog(LOG_INFO, "Joiner commission %s (psk: %s) (provisioning_url %s)", mAction ? "start" : "stop", mPsk, mProvisioningUrl);
```
abtink: Moving this before `if (action)`?

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+	mNextCommand = SpinelPackData(
+			SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_DATA_S),
+			SPINEL_PROP_THREAD_JOINER_COMMISSION,
+			frame.data(),
+			frame.size()
+			);
+
+	EH_SPAWN(&mSubPT, vprocess_send_command(event, args));
+	ret = mNextCommandRet;
+	check_noerr(ret);
```
abtink: how about `require_noerr(ret, on_error)` instead here?
`check_noerr()`  will add a print and continue (commonly used to indicate it's ok if this check fail).

src/ncp-spinel/SpinelNCPControlInterface.cpp
librasungirl: Existing `attach()` has both `NET_IF_UP` and `NET_STACK_UP`, however the interface has been up during previous commissioning process. No atomic `NET_IF_UP` and  `NET_STACK_UP` in existing implementation, that's why here introduces the `joiner_attach()` to avoid redundant `NET_IF_UP` operation.
I don't have strong thoughts on this, could reuse `attach()` if you think it's fine with the redundant `NET_IF_UP` operation.

src/ncp-spinel/SpinelNCPControlInterface.cpp
librasungirl: forgot to remove the debugging information output. will remove it.

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
librasungirl: will update.

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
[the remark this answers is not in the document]
librasungirl: will update.

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
librasungirl: will update as your proposal. Thanks.

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
```diff
@@ -0,0 +1,182 @@
...
+				SPINEL_FRAME_PACK_CMD_PROP_VALUE_SET(SPINEL_DATATYPE_BOOL_S),
+				SPINEL_PROP_NET_IF_UP,
+				false
+				);
+
+		EH_SPAWN(&mSubPT, vprocess_send_command(event, args));
+		ret = mNextCommandRet;
+		check_noerr(ret);
+	}
+
```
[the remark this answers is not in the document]
librasungirl: My concern here is that the time for commissioning process may depend on capability of specific platform and communication environment. Not sure what's the proper value to wait here. And we may also want to query the joiner state during the commissioning process. So the design here is to only trigger the commissioning process without caring about the results, and we can poll the status (ASSOCIATING, COMMISSIONED or CREDENTIALS_NEEDED) if we want to know the progress or results.
Thoughts?

src/ncp-spinel/SpinelNCPControlInterface.cpp
abtink: Thanks. Let's keep the `joiner_attach`.

src/ncp-spinel/SpinelNCPTaskJoinerCommission.cpp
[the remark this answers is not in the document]
abtink: Got it. Thanks. Sounds good.

librasungirl:
Rebased and updated. Please help to review it.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -3896,6 +3907,15 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
 			}
 			mResetIsExpected = false;
 			return;
+		} else if ((status >= SPINEL_STATUS_JOIN__BEGIN) && (status <= SPINEL_STATUS_JOIN__END)) {
+			if (status == SPINEL_STATUS_JOIN_SUCCESS)
+			{
```
abtink: nit: wpantund generally follows `{` on same line as the control command.
```
if (status == SPINEL_STATUS_JOIN_SUCCESS) {
    change_name_state(COMMISSIONED);
} else {
  ...
}
````

File: src/ncp-spinel/SpinelNCPTaskJoinerCommissioning.cpp [hunk 11]
```diff
@@ -0,0 +1,159 @@
...
+using namespace nl;
+using namespace nl::wpantund;
+
+nl::wpantund::SpinelNCPTaskJoinerCommissioning::SpinelNCPTaskJoinerCommissioning(
+	SpinelNCPInstance* instance,
+	CallbackWithStatusArg1 cb,
+	bool action,
+	const char *psk,
+	const char *provisioning_url
+	):	SpinelNCPTask(instance, cb), mAction(action),
```
abtink: nit: extra tab/space here

src/ncp-spinel/SpinelNCPTaskJoinerCommissioning.cpp
```diff
@@ -0,0 +1,159 @@
...
+	mPsk(psk), mProvisioningUrl(provisioning_url),
+	mLastState(instance->get_ncp_state())
+{
+}
+
+void
+nl::wpantund::SpinelNCPTaskJoinerCommissioning::finish(int status, const boost::any& value)
+{
+	if (mAction && status != kWPANTUNDStatus_Ok)
+	{
```
abtink: nit: `{` on same line as `if`

src/ncp-spinel/SpinelNCPTaskJoinerCommissioning.cpp
```diff
@@ -0,0 +1,159 @@
...
+	// line makes sure that we don't start processing this task
+	// until it is properly scheduled. All tasks immediately receive
+	// the initial `EVENT_STARTING_TASK` event, but further events
+	// will only be received by that task once it is that task's turn
+	// to execute.
+	EH_WAIT_UNTIL(EVENT_STARTING_TASK != event);
+
+	if (mAction)
+	{
+		mInstance->change_ncp_state(ASSOCIATING);
```
abtink: Let's save the `mLastState` here before changing state (since it may have changed from the time we save it from the constructor)

`mLastState = mInstance->get_ncp_state();`

src/ncp-spinel/SpinelNCPTaskJoinerCommissioning.cpp
```diff
@@ -0,0 +1,159 @@
...
+	// The first event to a task is EVENT_STARTING_TASK. The following
+	// line makes sure that we don't start processing this task
+	// until it is properly scheduled. All tasks immediately receive
+	// the initial `EVENT_STARTING_TASK` event, but further events
+	// will only be received by that task once it is that task's turn
+	// to execute.
+	EH_WAIT_UNTIL(EVENT_STARTING_TASK != event);
+
+	if (mAction)
+	{
```
abtink: nit: `{` on same line as `if`?

src/wpanctl/tool-cmd-joiner.c
```diff
@@ -0,0 +1,363 @@
...
+	const char* provisioning_url = NULL;
+	int psk_len = 0;
+	int provisioning_url_len = 0;
+	const char* property_joiner_state = kWPANTUNDProperty_ThreadJoinerState;
+	dbus_bool_t action = false;
+
+	dbus_error_init(&error);
+
+	if (argc == 1)
+	{
```
abtink: nit: `{` on same line as `if`

PR #331, opened 2018-07-16
Remove using DBUS_BUS_STARTER

bukepo:
Actually I'm not sure if this is the right cure, but we do encountered issues on raspberry PI desktop environment that either wpanctl or web gui cannot find the right dbus service of wpantund. Using the system dbus, it works as expected.

### DIFF
--- a/src/ipc-dbus/DBUSIPCServer.cpp
+++ b/src/ipc-dbus/DBUSIPCServer.cpp
@@ -65,13 +65,7 @@ get_dbus_connection()
 	if (connection == NULL) {
 		syslog(LOG_DEBUG, "Getting DBus connection");
 
-		connection = dbus_bus_get(DBUS_BUS_STARTER, &error);
-
-		if (!connection) {
-			dbus_error_free(&error);
-			dbus_error_init(&error);
-			connection = dbus_bus_get(DBUS_BUS_SYSTEM, &error);
-		}
+		connection = dbus_bus_get(DBUS_BUS_SYSTEM, &error);
 
 		require_string(connection != NULL, bail, error.message);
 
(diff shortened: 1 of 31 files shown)


bukepo:
@abtink any comments here?

abtink:
@bukepo  I am concerned about possible side-effect such a change may have on different platforms (to be honest, I am also not super familiar with how it impacts behavior).
Here is what I suggest:
- How about if we bring in this change with a build-time config flag.
- We can define a common function `get_dbus_connection()` in `wpanctl-utils.h` and `wpanctl-utils.c` which all other wpanctl files then use to get a dbus connection
- We can  have it like this so we can use `WPANCTL_USE_SYSTEM_DBUS` to force it to directly use the `DBUS_BUS_SYSTEM`. This way the default behavior stays as before
```
#ifdef WPANCTL_USE_SYSTEM_DBUS
DBusConnection *get_dbus_connection(DBusError *error)
{
	return dbus_bus_get(DBUS_BUS_SYSTEM, error);
}
#else
DBusConnection *get_dbus_connection(DBusError *error)
{
	DBusConnection *connection = NULL;

	connection = dbus_bus_get(DBUS_BUS_STARTER, error);

	if (!connection) {
		dbus_error_free(error);
		dbus_error_init(error);
		connection = dbus_bus_get(DBUS_BUS_SYSTEM, error);
	}

	return connection;
}
#endif
```

bukepo:
@abtink Thanks for providing the solution, however as @darconeous suggested, it seems we will not connect to session bus which is usually used in desktop environment for user login sessions. 
 This wpantund is a system-wide service, it's not bound to a user.
Besides, I have applied the changes to new commands.

abtink:
@bukepo sounds good to me. I merged https://github.com/openthread/wpantund/pull/340 a moment ago, which created a conflict with this. Can you please help rebase this again then we merge it.  Thanks.

bukepo:
@abtink rebased to master, thank you for this notification.

PR #334, opened 2018-07-24
Allow multiple instances of same on-mesh prefix to be tracked separately

abtink:
This commit change how the on-mesh prefix list is maintained by
`wpantund` to allow for instances of same prefix added by different
nodes within the Thread network (possibly with different flags) to be
tracked individually.  This commit also changes the behavior when an
IPv6 address is added with an existing prefix (e.g., prefix already
added by another node). In such case the prefix will be re-added by
the node adding the new IPv6 address.


abtink:
Tested internally.

PR #338, opened 2018-08-06
Fix the `encode_data_into_string()` implementation

abtink:
This commit fixes an issue in `encode_data_into_string()` where
the last byte may be missed in string representation.


abtink:
Verified internally.

PR #339, opened 2018-08-07
Enter ISOLATED state after resume on a sleepy child

abtink:
This commit changes the wpantund state after a reset/resume such that
if the device is a sleepy child, it is starts from `ISOLATED` state
("associated:no-parent") instead of `ASSOCIATING` state. This is
handled from `PROP_NET_STACK_UP` property update by checking if the
device is commissioned AND the thread mode indicates that the device
is sleepy. This commit also ensures that if thread mode is changed
while in either `ISOLATED` or `ASSOCIATING` states, the state is
switched accordingly based on the new mode and whether it is sleepy
or not.

The reason for this change is to ensure wpantund allows host sleep on
a sleepy child with no parent after an NCP reset. Since the
`ASSOCIATING` state is considered a transient state, wpantund blocks
host sleep while in this state. However, host sleep is correctly
allowed while in `ISOLATED` state.


mszczodrak:
LGMT

PR #343, opened 2018-08-17
Enhance android makefile

bukepo:
* enhanced android makefile to deal with new files in `src/ncp-spinel/`
* added android check in travis so that we know the changes will not block build on Android.


.travis/script.sh
```diff
@@ -18,11 +18,775 @@
...
+
+/* Define to \`__inline__' or \`__inline' if that's what the C compiler
+   calls it, or to nothing if 'inline' is not supported under any name.  */
+#ifndef __cplusplus
+/* #undef inline */
+#endif
+EOF
+
+    cat > "${DBUS_SOURCE}/dbus/dbus-arch-deps.h" <<EOF
+/* -*- mode: C; c-file-style: "gnu" -*- */
```
abtink: How about adding a copy of this as a file somewhere in "third_party" folder?

.travis/script.sh
```diff
@@ -18,11 +18,775 @@
...
+# Returns:
+#   None
+#######################################
+android_prepare_dbus()
+{
+    DBUS_SOURCE=dbus-1.4.26
+    wget "https://dbus.freedesktop.org/releases/dbus/${DBUS_SOURCE}.tar.gz"
+    tar xvf "${DBUS_SOURCE}.tar.gz"
+    cat > "${DBUS_SOURCE}/config.h" <<EOF
+/* config.h.  Generated from config.h.in by configure.  */
```
abtink: How about adding a copy of this as a file instead of embedded in the script?

Android.mk
```diff
@@ -85,6 +85,8 @@ LOCAL_LDFLAGS := \
 $(LOCAL_PATH)/src/version.c: $(LOCAL_PATH)/src/version.c.in
 	sed 's/SOURCE_VERSION/"$(LOCAL_PRIVATE_SOURCE_VERSION)"/' < $< > $@
 
+NCP_SPINEL_SRC_FILES := $(wildcard $(LOCAL_PATH)/src/ncp-spinel/*.cpp) $(wildcard $(LOCAL_PATH)/src/ncp-spinel/*.c)
```
abtink: We don't need `*.h` files?

bukepo: Android.mk is different from automake. We don't need list `*.h` files. I also removed other header files.

.travis/script.sh
bukepo: As you see there are some other files generated in this script, all of them are just used for travis test. Some of these files are tricks to cheat the android build system. I prefer keeping these files generated at runtime.

bukepo:
@abtink I removed the generated files from the script.

PR #352, opened 2018-09-27
Re-initialize NCP on driver start after a stop

abtink:

1) It adds new code to restart the main proptothread from
   `SpinelNCPInstance::vprocess_event()` when driver is enabled again
   after a driver disable. This ensures internal state stays in sync
   with NCP and that the network is resumed if `mAutoResume` feature
   is enabled.

2) The code changes how the interface is taken offline when driver
   is disabled.
   It updates `NCPInstanceBase::handle_ncp_state_change()` to ensure
   when driver is disabled causing a transition to `COMMISSIONED`
   state even if `mAutoResume` is enabled, the network interface(s)
   are taken down.
   It also removes the call from `vprocess_disabled()` protothread
   to bring down the interface.

### DIFF
--- a/src/wpantund/NCPInstanceBase.cpp
+++ b/src/wpantund/NCPInstanceBase.cpp
@@ -1292,6 +1292,11 @@ NCPInstanceBase::handle_ncp_state_change(NCPState new_ncp_state, NCPState old_nc
 	} else if (ncp_state_is_interface_up(old_ncp_state)
 				&& (new_ncp_state == COMMISSIONED)
 				&& mAutoResume
+				// The `mEnabled` check covers the case where driver is disabled
+				// causing a transition to COMMISSIONED state. In that case
+				// we want to fall to "InterfaceUp -> InterfaceDown (General Case)"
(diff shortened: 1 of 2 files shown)


abtink:
The failing `toranj` test (`test-100-mcu-power-state.py`) is expected due to the change in state transitions after `"Daemon:Enabled"` value change 
Submitted  https://github.com/openthread/openthread/pull/3103 would relax the state checks.

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -1292,6 +1292,11 @@ NCPInstanceBase::handle_ncp_state_change(NCPState new_ncp_state, NCPState old_nc
 	} else if (ncp_state_is_interface_up(old_ncp_state)
 				&& (new_ncp_state == COMMISSIONED)
 				&& mAutoResume
+				// The `mEnaled` check covers the case where driver is disabled
```
jwhui: typo: `mEnabled`

PR #354, opened 2018-10-01
only remove slaac address if no available border routers

librasungirl:
fix 7.1.7 OTBR DUT Leader

### DIFF
--- a/src/wpantund/NCPInstanceBase-Addresses.cpp
+++ b/src/wpantund/NCPInstanceBase-Addresses.cpp
@@ -787,6 +787,27 @@ NCPInstanceBase::lookup_address_for_prefix(struct in6_addr *address, const struc
 	return false;
 }
 
+bool NCPInstanceBase::has_slaac_on_mesh_prefix(Origin origin, const IPv6Prefix &prefix)
+{
+	bool rval = false;
+	std::multimap<IPv6Prefix, OnMeshPrefixEntry>::iterator iter;
+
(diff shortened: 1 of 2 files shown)


librasungirl:
@jwhui , @abtink , please help to have a review. Thanks.

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -874,9 +895,12 @@ NCPInstanceBase::on_mesh_prefix_was_removed(Origin origin, const struct in6_addr
 		if (lookup_address_for_prefix(&address, prefix.get_prefix(), prefix.get_length())
 				&& entry.is_slaac() && entry.is_on_mesh() && prefix.get_length() == 64
 		) {
+			if (mOnMeshPrefixes.count(prefix) == 0 || !has_slaac_on_mesh_prefix(origin, prefix))
```
abtink: Not sure if we should care where the origin is here in the call to `has_slaac_on_mesh_prefix`.
I think if we do have a matching on-mesh prefix with slaac from any origin, we should keep the SLAAC address.

Another point I noticed, if user explicitly added an IPv6 address which matched the prefix (which is actually what weave does), then we should not assume it's slaac address and remove it.

jwhui: Let's fix this up in a separate PR.

abtink: I am working on PR to supplement this.

PR #366, opened 2018-11-05
Enabled latency test for NCP

LeZhang2016:
This feature need to enable time sync in advance on NCP.

In this commit, a command line "performance" is added into wpanctl,
user could use this command to set:
1. sending node or receiving node
2. UDP payload size
3. IPv6 destination
In addition, added two "get handlers" for latency and hoplimit, which
can be obtained from the receiving node after executing "performance"
command.

This PR is related to OT PR#3269
https://github.com/openthread/openthread/pull/3269


abtink:
@LeZhang2016 , I am closing older PRs. Please feel free to edit (fix conflicts) and re-submit.

PR #367, opened 2018-11-05
Notify of network time updates and timestamp incoming network time messages with system mono time

josephn123:
This pull request covers the following:

- After this openthread PR (https://github.com/openthread/openthread/pull/3238), the NCP will notify the host of network time and network time status changes.  This pull request includes forwarding this notification as a dbus signal to clients.
- Timestamp network time/status messages with the system mono time using boost steady_clock. This is so the client can correct the received network time by eliminating the time taken from reception by wpantund to the point it receives the message over dbus.

Most of the files in this PR are adding the required header files for boost::chrono.


josephn123:
Build is failing because `libboost_system_dev` and `libboost_chrono_dev` are not present on the CI build agent. Not sure how I can change this?

josephn123:
The corresponding change on openthread has been merged: https://github.com/openthread/openthread/commit/4f99bcfaf3c1a561d59cc40aa1570ab23c1bace1

abtink:
How about if we have the client (dbus signal handler) to do the time-stamping? This removes the need for boost:chrono (which can add a dependency on wpantund). The DBUS signal delay should be small compared to NCP delay (spinel preparaing, queueing, uart/spi interface, ect).

josephn123:
> How about if we have the client (dbus signal handler) to do the time-stamping? This removes the need for boost:chrono (which can add a dependency on wpantund). The DBUS signal delay should be small compared to NCP delay (spinel preparaing, queueing, uart/spi interface, ect).

Some rough figures I've seen is about 16ms delay from wpantund receiving the packet to DBus-Java receiving the signal. Granted that DBus-Java may (or may not be) slower than other implementations. I haven't actually got any figures yet for the delay you mention but 16ms is still fairly significant in my opinion, particularly as I will be working on eliminating that delay (by having the NCP trigger a host GPIO pin when sending the time sync message and having the host correct the subsequently received time sync message based upon the recorded monotonic time of when the GPIO interrupt was fired).

abtink:
I did not expect 16ms dbus signal delay. 

One concern I have is that this change seem to add dependency of boost::chrono to `wpantund`. This adds a limitation on the platforms where `wpantund` is used even if the corresponding NCP does not support/enable the time sync feature. Would be good if we can add this dependency in a build/configure option so it can be enabled/disabled at build time if not needed. 

About travis CI missing `libboost_system_dev` and `libboost_chrono_dev`,  I think you can update the [`.travis.yml`](https://github.com/openthread/wpantund/blob/master/.travis.yml#L3) file and add them to the packages list.

josephn123:
Updated with optional `configure` flag `--enable-append-network-time-received-timestamp`

src/wpantund/NCPControlInterface.h
```diff
@@ -284,6 +284,12 @@ class NCPControlInterface {
 
 	boost::signals2::signal<void(const EnergyScanResultEntry&)> mOnEnergyScanResult;
 
+public:
+	// ========================================================================
+	// Network time-related Member Functions
+
+	boost::signals2::signal<void(const NetworkTimeUpdate&)> mOnNetworkTimeUpdate;
```
abtink: How about changing this signal format to use `ValueMap` as input param instead?
```
boost::signals2::signal<void(const ValueMap &)> mOnNetworkTimeUpdate;
```
This will be more flexible and would help later with preparing the DBus message.

src/wpantund/NCPInstanceBase.cpp
```diff
@@ -1371,6 +1371,15 @@ NCPInstanceBase::reset_tasks(wpantund_status_t status)
 {
 }
 
+// ----------------------------------------------------------------------------
+// MARK: Network Time Update
+
+void
+NCPInstanceBase::handle_network_time_update(const NetworkTimeUpdate& update)
```
abtink: Same here. Suggest changing param to be `const ValueMap &update`.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -4568,6 +4610,19 @@ SpinelNCPInstance::handle_ncp_spinel_value_is(spinel_prop_key_t key, const uint8
...
+	} else if (key == SPINEL_PROP_THREAD_NETWORK_TIME) {
+		boost::any value;
+		NetworkTimeUpdate update;
+
+		if (unpack_thread_network_time(value_data_ptr, value_data_len, value, true) == kWPANTUNDStatus_Ok && 
+			extract_thread_network_time(value, update)) {
+			
+			syslog(LOG_INFO, "[-NCP-]: Network time update: network_time:%lu, status:%d", update.mNetworkTime, update.mStatus);
+
+			handle_network_time_update(update);
```
abtink: I think if we change the `handle_network_time_update()` to use `const ValueMap &` as input parameter, it would make this simpler as we won't need the `extract_thread_network_time()` here and can directly pass the value map.

For logging we can 
- either use the valuemap and check the entries under the keys `kWPANTUNDValueMapKey_TimeSync_Time` and `kWPANTUNDValueMapKey_TimeSync_Status`, 
- or call `unpack_thread_network_time()` with `as_val_map` parameter as `false` and log the resulting string.

src/ipc-dbus/DBusIPCAPI_v1.cpp
```diff
@@ -321,6 +329,69 @@ DBusIPCAPI_v1::received_beacon(NCPControlInterface* interface, const WPAN::Netwo
 	dbus_message_unref(signal);
 }
 
+static void
+ipc_append_network_time_update_dict(
+	DBusMessageIter *iter, const NetworkTimeUpdate& network_time_update)
+{
+	DBusMessageIter dict;
+
+	dbus_message_iter_open_container(
```
abtink: If we change the parameter to be `const ValueMap &`, it should make building the DBus message simpler.
We can use `append_any_to_dbus_iter()` here and it would take care of building the DBus dictionary.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -45,6 +45,7 @@
 #define kWPANTUND_Whitelist_RssiOverrideDisabled    127
 #define kWPANTUND_SpinelPropValueDumpLen            8
 
+
```
abtink: nit: extra newline

abtink:
@josephn123, looks good overall. Thanks for submitting the PR and adding `--enable-append-network-time-received-timestamp` 

I added some suggestion above.

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -1252,6 +1253,9 @@ unpack_thread_network_time(const uint8_t *data_in, spinel_size_t data_len, boost
 			entry.clear();
```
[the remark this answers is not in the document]
josephn123: I agree with your conclusion here. However, changing this will affect existing users of wpantund who may have written dbus client code to expect this structure.

Example, when running `getprop TimeSync:ThreadNetworkTime:AsValMap` from `wpanctl`:

Before:

```
TimeSync:ThreadNetworkTime:AsValMap = [
	[
		"ThreadNetworkTime" => 0x000000000030AD25
		"TimeSyncReceivedMonoTimeUs" => 0x000000227F5DF606
		"TimeSyncStatus" => 1
	]
]
```

After:
```
TimeSync:ThreadNetworkTime:AsValMap = [
	"ThreadNetworkTime" => 0x000000000305724B
	"TimeSyncReceivedMonoTimeUs" => 0x000000227A871491
	"TimeSyncStatus" => 1
]
```

src/ipc-dbus/DBusIPCAPI_v1.cpp
josephn123: Great suggestion, thanks. This has also greatly simplified my DBus client code. Pull request updated

josephn123:
What are your thoughts on adding an additional entry to the `env` section in `.travis.yml`?

```
BUILD_MAKEPATH=build BUILD_MAKEARGS="distcheck" TAR=bsdtar BUILD_CONFIGFLAGS=--with-boost=internal --enable-append-network-time-received-timestamp
```

Advantage: So we can be sure this still builds in the future and links with `boost_chrono` and `boost_system`
Disadvantage: CI build time is longer for everyone

src/ncp-spinel/SpinelNCPInstance.cpp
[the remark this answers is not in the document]
abtink: I see your point. I agree, let's keep it as is.

abtink:
LGTM.  Thanks @josephn123 
Good idea about adding a new entry in travis.

PR #369, opened 2018-11-29
Fix warnings on build

ramonhpr:
This patch set setup the configure.ac to generate the headers with _GNU_SOURCE macro.
Also, some annoying warnings about the string_util include.


abtink:
Thanks for your PR and contributions. Can you please check if you can sign the CLA.

PR #376, opened 2019-01-23
[ip6] handle adding SLAAC inside OpenThread

bukepo:
This is a follow up of openthread/openthread#3287


src/wpantund/wpan-properties.h
```diff
@@ -51,7 +51,6 @@
 
 #define kWPANTUNDProperty_DaemonIPv6AutoUpdateIntfaceAddrOnNCP  "Daemon:IPv6:AutoUpdateInterfaceAddrsOnNCP"
 #define kWPANTUNDProperty_DaemonIPv6FilterUserAddedLinkLocal    "Daemon:IPv6:FilterUserAddedLinkLocal"
-#define kWPANTUNDProperty_DaemonIPv6AutoAddSLAACAddress         "Daemon:IPv6:AutoAddSLAACAddress"
```
abtink: We need this property and capability to not add SLAAC addresses.
(this is being used internally)

PR #386, opened 2019-03-08
Add ALOC service support

josephn123:
Adds support for ALOC services: https://github.com/openthread/openthread/issues/3651


josephn123:
Should wpantund keep track of services added in case the NCP is reset?

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -5080,6 +5251,62 @@ SpinelNCPInstance::remove_multicast_address_on_ncp(const struct in6_addr &addr,
...
 }
 
+void
+SpinelNCPInstance::add_service_on_ncp(uint32_t enterprise_number, 
+	const uint8_t *service_data, unsigned int service_data_len, bool stable, 
+	const uint8_t *server_data, unsigned int server_data_len, CallbackWithStatus cb)
+{
+	SpinelNCPTaskSendCommand::Factory factory(this);
+
+	syslog(LOG_NOTICE, "Adding service with enterprise number:%u to NCP", enterprise_number);
```
abtink: Suggestion: how about checking the capability for `SPINEL_CAP_THREAD_SERVICE` and if it's not supported, invoking callback `cb` with error?

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -281,6 +281,42 @@ SpinelNCPControlInterface::remove_external_route(
...
+	CallbackWithStatus cb
+) {
+	require_action(service_data != NULL, bail, cb(kWPANTUNDStatus_InvalidArgument));
+	require_action(service_data_len > 0, bail, cb(kWPANTUNDStatus_InvalidArgument));
+	require_action(server_data != NULL, bail, cb(kWPANTUNDStatus_InvalidArgument));
+	require_action(server_data_len > 0, bail, cb(kWPANTUNDStatus_InvalidArgument));
+	require_action(mNCPInstance->mEnabled, bail, cb(kWPANTUNDStatus_InvalidWhenDisabled));
+
+	mNCPInstance->add_service(enterprise_number, service_data, service_data_len, stable, server_data, server_data_len, cb);
+
```
abtink: We can consider directly implementing the spinel exchange here from the `SpinelNCPControlInterface` (e.g. similar to `commissioner_add_joiner`). With this mode we can remove the `NCPInstanceBase` and `SpinelNCPInstance` related APIs (`add_service_on_ncp` and `add_service`) which would simplify the logic.

For addresses/routes/prefixes we go through the `NCPInstanceBase` class since it implements the mechanism to save the user-originated  entries (from `NCPInstanceBase-Addresses`)  and ensures to re-add them on NCP on NCP reset/reboot.

This PR does not currently save the "service" entries in `wpantund`. Not sure if we would need this in future. If we do, I'd say we should keep the current flow (go through the NCPInstanceBase), otherwise we can directly implement the logic in `SpinelNCPControlInterface`.

Thoughts?

src/ncp-spinel/SpinelNCPControlInterface.h
```diff
@@ -163,6 +163,23 @@ class SpinelNCPControlInterface : public NCPControlInterface, public NCPMfgInter
 		CallbackWithStatus cb = NilReturn()
 	);
 
+	virtual void add_service(
+		uint32_t enterprise_number, 
+		const uint8_t *service_data, 
+		unsigned int service_data_len,
```
abtink: Suggestion: How about using `Data` type for parameters service and server data?  `const Data &service_data`?

src/ncp-spinel/SpinelNCPControlInterface.cpp
josephn123: Think it could be advantageous to restore the service entries by saving it in wpantund. At least I have noticed most configuration is saved and restored, whether that is in wpantund or on the NCP flash itself. Otherwise we push the complexity in to the user's supervisor application (connected via DBus) to monitor resets and re-add services.

I will look in to adding this functionality.

josephn123: Done

src/ncp-spinel/SpinelNCPInstance.cpp
josephn123: Done

src/ncp-spinel/SpinelNCPControlInterface.h
josephn123: Done

josephn123:
@abtink - any further suggestions?

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -557,6 +558,12 @@ SpinelNCPInstance::get_supported_property_keys()const
 		properties.insert(kWPANTUNDProperty_TimeSync_XtalThreshold);
 	}
 
+	if (mCapabilities.count(SPINEL_CAP_THREAD_SERVICE))
+	{
```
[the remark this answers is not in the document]
josephn123: Done

src/ncp-spinel/SpinelNCPInstance-Protothreads.cpp
```diff
@@ -470,6 +470,7 @@ SpinelNCPInstance::vprocess_init(int event, va_list args)
 				{ SPINEL_PROP_NET_ROLE, 0 },
 				{ SPINEL_PROP_SLAAC_ENABLED, SPINEL_CAP_SLAAC },
 				{ SPINEL_PROP_RCP_VERSION , SPINEL_CAP_POSIX_APP },
+				{ SPINEL_PROP_SERVER_SERVICES, 0 },
```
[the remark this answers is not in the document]
josephn123: Done

PR #387, opened 2019-03-13
NCP supporting SLAAC

abtink:
This commit adds new wpan properties to control SLAAC behavior on
NCP (if supported). It adds "OpenThread:SLAAC:Enabled" wpan property
to check status of SLAAC on NCP/OpenThread. The new code also
ensures to disable SLAAC address addition by `wpantund` if NCP
already supports SLAAC.


abtink:
Verified internally.

PR #390, opened 2019-03-22
Add group to allow sending d-bus messages to wpantund

### LINKED ISSUE #389: [d-bus] at_console is deprecated.
**Describe the bug**
at_console is deprecated, and not supported under ChromeOS.

**Expected behavior**
remove this from d-bus configuration file.

**Additional context**
I think we should use group to allow clients.

bukepo:
This fixes #389 . I didn't remove `at_console` section to avoid breaking existing projects.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,6 +132,25 @@ AC_ARG_ENABLE(
 )
 
 #
+# Service group
+#
+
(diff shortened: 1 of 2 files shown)


bukepo:
@abtink Do you have any comments here?

PR #415, opened 2019-05-24
Log neighbor/router entry added/removed events from NCP

abtink:
---
This is related to https://github.com/openthread/openthread/pull/3838

Example of how the  entries are loggd:
```
wpantund[60873]: [-NCP-]: Neighbor(Router) entry added: 1B007E862ACBA011, RLOC16:0f82, LQIn:0, AveRssi:0, LastRssi:0, Age:51200, LinkFC:196875, MleFC:327680, IsChild:yes, RxOnIdle:no, FTD:yes, SecDataReq:no, FullNetData:yes
wpantund[60873]: [-NCP-]: Neighbor(Router) entry added: 1B00428DC3508D20, RLOC16:78b6, LQIn:0, AveRssi:0, LastRssi:0, Age:36864, LinkFC:327947, MleFC:917504, IsChild:yes, RxOnIdle:no, FTD:yes, SecDataReq:no, FullNetData:yes
```

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -5100,6 +5100,16 @@ SpinelNCPInstance::handle_ncp_spinel_value_inserted(spinel_prop_key_t key, const
 			syslog(LOG_INFO, "[-NCP-]: ChildTable entry added: %s", child_entry.get_as_string().c_str());
 		}
 
+	} else if (key == SPINEL_PROP_THREAD_NEIGHBOR_TABLE) {
+		SpinelNCPTaskGetNetworkTopology::TableEntry neighbor_entry;
+		int status;
+
+		status = SpinelNCPTaskGetNetworkTopology::parse_neighbor_entry(value_data_ptr, value_data_len, neighbor_entry);
+
+		if (status == kWPANTUNDStatus_Ok) {
+			syslog(LOG_INFO, "[-NCP-]: Neighbor(Router) entry added: %s", neighbor_entry.get_as_string().c_str());
+		}
+
 	} else if (key == SPINEL_PROP_MESHCOP_COMMISSIONER_ENERGY_SCAN_RESULT) {
 		spinel_ssize_t len;
 		uint32_t channel_mask;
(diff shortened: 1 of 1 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -5100,6 +5100,16 @@ SpinelNCPInstance::handle_ncp_spinel_value_inserted(spinel_prop_key_t key, const
...
 		}
 
+	} else if (key == SPINEL_PROP_THREAD_NEIGHBOR_TABLE) {
+		SpinelNCPTaskGetNetworkTopology::TableEntry neighbor_entry;
+		int status;
+
+		status = SpinelNCPTaskGetNetworkTopology::parse_neighbor_entry(value_data_ptr, value_data_len, neighbor_entry);
+
+		if (status == kWPANTUNDStatus_Ok) {
+			syslog(LOG_INFO, "[-NCP-]: Neighbor(Router) entry added: %s", neighbor_entry.get_as_string().c_str());
```
josephn123: Use after free: `syslog()` is receiving a pointer to an already freed c string because the returned `std::string` will no longer be in scope once `c_str()` returns.

Ensure the string remains in scope:
```
std::string neighbor_entry_str = neighbor_entry.get_as_string();
syslog(LOG_INFO, "[-NCP-]: Neighbor(Router) entry added: %s", neighbor_entry_str.c_str());
```

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -5175,6 +5185,16 @@ SpinelNCPInstance::handle_ncp_spinel_value_removed(spinel_prop_key_t key, const
...
 		}
 
+	} else if (key == SPINEL_PROP_THREAD_NEIGHBOR_TABLE) {
+		SpinelNCPTaskGetNetworkTopology::TableEntry neighbor_entry;
+		int status;
+
+		status = SpinelNCPTaskGetNetworkTopology::parse_neighbor_entry(value_data_ptr, value_data_len, neighbor_entry);
+
+		if (status == kWPANTUNDStatus_Ok) {
+			syslog(LOG_INFO, "[-NCP-]: Neighbor(Router) entry removed: %s", neighbor_entry.get_as_string().c_str());
```
josephn123: Same use after free issue as above

src/ncp-spinel/SpinelNCPInstance.cpp
abtink: @josephn123 , the existing model is actually correct and intentional. We do commonly use this pattern in `wpantund` and also in OpenThread (e.g., see the `ot::String` class in OpenThread). 

A temporarily created object stays in scope for the duration of its full enclosing statement which in this case would be the `syslog()` call. After `syslog()` returns the temp `std::string` object would be destroyed.

> All temporary objects are destroyed as the last step in evaluating the full-expression that (lexically) contains the point where they were created, and if multiple temporary objects were created, they are destroyed in the order opposite to the order of creation. This is true even if that evaluation ends in throwing an exception.
https://en.cppreference.com/w/cpp/language/lifetime

https://stackoverflow.com/questions/10006891/stdstringc-str-and-temporaries

josephn123: Thanks for the info, didn't know this. Thought I'd encountered the same issue one time for the reason I had described

Issue #189, opened 2017-06-14
Illegal Network Name causes wpantund to crash


DuaneEllis-TI:
From: https://github.com/openthread/openthread/issues/1902

Per discussion with @jwhui  in Openthread#1902, this should probably be handled within wpantund.
Hence I'm moving/opening the issue here.

[This is an edited version of that entry]

This actually crashes wpantund, but has roots in the mis-behaving remote device.

From the 'wpanctl' application, use the "scan" command to scan for networks. The code in wpantund (dbus, specifically) assumes, rightly or wrongly, that the network name is a UTF8 string.

the NCP transmits a beacon request.

In this scenario, the remote device answering the beacon request has a bug and is sending a bad or invalid string as the network name. That device should transmit: "OpenThread" - instead it transmits: "O\377\377\377\377hread" (Where: \377 = Octal for an 0xFF byte). It looks like something wrote a 32bit "-1" value over the top of something it should not have.

This goes over the air with the "0xff" bytes in place of a portion of the network name, up through the NCP and finally into Wpantund - where an assertion triggers this error:

```
process 592: arguments to dbus_message_iter_append_basic() were incorrect, assertion "_dbus_check_is_valid_utf8 (*string_p)" failed in file ../../dbus/dbus-message.c line 2676.
```

of course this should be fixed in the transmitting device, but the robustness principle states either the NCP, or WPANTUND ... something, somewhere shall not die a horrible death as it is right now.

=== 

-Duane.

darconeous:
Such strings used to be sanitized before letting them on DBus. I'm guessing that code was somehow lost in a refactor at some point.

Vyrastas:
I encountered the same issue, which for me was caused by a networkname of exactly 16 characters.  Create any Thread network with a networkname of 16 characters, then scan with wpanctl.  wpantund will crash:

wpantund[28027]: Scan -> Name:codelabacodelabb���z���z�, PanId:0xC0DE, Ch:11, Joinable:NO , XPanId:0xC0DE7AB5C0DE7AB5, HwAddr:0x62991E21E38AE4B5, RSSI:-39 , LQI:54 , ProtoId:3  , Version: 0, ShortAddr:0xFFFF
process 28027: arguments to dbus_message_iter_append_basic() were incorrect, assertion "_dbus_check_is_valid_utf8 (*string_p)" failed in file ../../../dbus/dbus-message.c line 2753.
This is normally a bug in some application using the D-Bus library.
[...]

PR #420, opened 2019-06-12
Update spinel files from OpenThread (#418)

abtink:
This commit update spinel source files from OpenThread at commit
dbc04a8a0. This commit brings in the following change from OpenThread
"[spinel] validate decoded/unpacked UTF8 string (#3889)" which
ensures a decoded string follows UTF8 encoding.


abtink:
This PR will address Issue https://github.com/openthread/wpantund/issues/189

PR #428, opened 2019-07-17
Wait till end of NCP init when starting a new command task

abtink:
This commit changes the `NCPTask<Command>` protothread such that at
the start we wait for the end of any ongoing NCP initialization.
This commit adds the condition `!mInstance->is_initializing_ncp()` in
addition to checking of ncp state. Note that ncp state from
`get_ncp_state()` itself may change during initialization while
driver is syncing with NCP.


abtink:
Not sure about failing immediately. It would be a big change in behavior. The current code handles situations where NCP itself may be reset (e.g., due to a crash/assert/etc).  If user has issued a command (waiting in the task queue), we wait for NCP to be re-initialized and then execute the command. So user won't need to be aware of the reset.

bukepo:
@abtink I agree.

PR #430, opened 2019-07-18
Make "reset" wait till end of NCP init

abtink:
This commit convert the `SpinelNCPControlInterface::reset()` API to
wait till end of NCP re-initilaization (instead of returning after
issuing the `RESET` command to NCP). This is realized by adding a
feature in `TaskSendCommand` to allow an optional "final check
handler" to be added to the command. If provided after the commands
are sent to NCP, the task will invoke the provided handler and waits
until the handler returns `true` (within the given timeout interval).

### DIFF
--- a/src/ncp-spinel/SpinelNCPControlInterface.cpp
+++ b/src/ncp-spinel/SpinelNCPControlInterface.cpp
@@ -129,6 +129,11 @@ SpinelNCPControlInterface::attach(CallbackWithStatus cb)
 	);
 }
 
+static bool check_ncp_finished_initalization(SpinelNCPInstance *instance)
+{
+	return !ncp_state_is_initializing(instance->get_ncp_state()) && !instance->is_initializing_ncp();
+}
+
 void
 SpinelNCPControlInterface::reset(CallbackWithStatus cb)
 {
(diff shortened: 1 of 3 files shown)


abtink:
Here is wpantund logs (all log level enabled) with the change in this PR showing the DBus response to "reset" is sent after NCP initialization is done:

```
wpantund[132255]: Inbound DBus message for INTERFACE "Reset" from ":1.178489"
wpantund[132255]: [->NCP] (11) RESET
wpantund[132255]: [NCP->] (0) PROP_VALUE_IS(LAST_STATUS) [70]
wpantund[132255]: [-NCP-]: Last status (RESET_POWER_ON, 112)
wpantund[132255]: [-NCP-]: NCP was reset (RESET_POWER_ON, 112)
wpantund[132255]: State change: "offline" -> "uninitialized"
wpantund[132255]: Initializing NCP
...
...
...
wpantund[132255]: Resetting interface(s). . .
wpantund[132255]: Resetting interface wpan2. . .
wpantund[132255]: Taking interface wpan2 offline. . .
wpantund[132255]: DBus Sending Association State Changed to offline
wpantund[132255]: property_get_value: key: "Daemon:Enabled"
wpantund[132255]: Finished initializing NCP
wpantund[132255]: Sending DBus response for "Reset" to ":1.178489"
```

src/ncp-spinel/SpinelNCPControlInterface.cpp
```diff
@@ -129,6 +129,11 @@ SpinelNCPControlInterface::attach(CallbackWithStatus cb)
 	);
 }
 
+static bool check_ncp_finished_initalization(SpinelNCPInstance *instance)
+{
+	return !ncp_state_is_initializing(instance->get_ncp_state()) && !instance->is_initializing_ncp();
```
gjc13: Will there be such case that the final_check is called even before the ncp instance sets its state to initializing and starts to reset the device? Shall we also check ncp is not in a "working" state?

PR #435, opened 2019-09-11
Add thread coex enable/disable properity

zhanglongxia:
This commit adds "SPINEL_PROP_RADIO_COEX_ENABLE" to the list of spinel properties and
add "kWPANTUNDProperty_NCPCoexEnable" wpan properties to enable/disable the radio coex.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -452,7 +452,6 @@ SpinelNCPInstance::get_supported_property_keys()const
 	properties.insert(kWPANTUNDProperty_NCPExtendedAddress);
 	properties.insert(kWPANTUNDProperty_NCPCCAFailureRate);
 	properties.insert(kWPANTUNDProperty_NCPCapabilities);
-	properties.insert(kWPANTUNDProperty_NCPCoexMetrics);
 
 	if (mCapabilities.count(SPINEL_CAP_ROLE_SLEEPY)) {
 		properties.insert(kWPANTUNDProperty_NCPSleepyPollInterval);
@@ -567,6 +566,11 @@ SpinelNCPInstance::get_supported_property_keys()const
 	if (mCapabilities.count(SPINEL_CAP_THREAD_SERVICE)) {
 		properties.insert(kWPANTUNDProperty_ThreadServices);
 		properties.insert(kWPANTUNDProperty_ThreadLeaderServices);
+	}
+
+	if (mCapabilities.count(SPINEL_CAP_RADIO_COEX)) {
+		properties.insert(kWPANTUNDProperty_NCPCoexEnable);
+		properties.insert(kWPANTUNDProperty_NCPCoexMetrics);
 	}
 
 	{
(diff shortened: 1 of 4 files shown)


src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -3138,6 +3141,9 @@ SpinelNCPInstance::regsiter_all_set_handlers(void)
 	register_set_handler_spinel(
 		kWPANTUNDProperty_ThreadRouterDowngradeThreshold,
 		SPINEL_PROP_THREAD_ROUTER_DOWNGRADE_THRESHOLD, SPINEL_DATATYPE_UINT8_C);
+	register_set_handler_spinel(
+		kWPANTUNDProperty_NCPCoexEnable,
+		SPINEL_PROP_RADIO_COEX_ENABLE, SPINEL_DATATYPE_BOOL_C);
```
abtink: Two suggestions: 
-  I suggested we add a spinel cap for COEX, which we can then use here.
- I also think we want to persist this property in wpantund (if the user sets it and afterwards NCP gets reset, we remember the value in `wpantund` and set it back on NCP during its initialization).

To change this to persist and perform check capability, we just need to use the method `register_set_handler_capability_spinel_persist()` instead to register the property

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -1964,6 +1964,9 @@ SpinelNCPInstance::regsiter_all_get_handlers(void)
 	register_get_handler_spinel_simple(
 		kWPANTUNDProperty_OpenThreadLogLevel,
 		SPINEL_PROP_DEBUG_NCP_LOG_LEVEL, SPINEL_DATATYPE_UINT8_S);
+	register_get_handler_spinel_simple(
+		kWPANTUNDProperty_NCPCoexEnable,
+		SPINEL_PROP_RADIO_COEX_ENABLE, SPINEL_DATATYPE_BOOL_S);
```
abtink: Same suggestion here to register with spinel capability check for coex (using `register_get_handler_capability_spinel_simple()`)

We can also add same capability check to metrics property

src/ncp-spinel/SpinelNCPInstance.cpp
zhanglongxia: Have updated.

src/ncp-spinel/SpinelNCPInstance.cpp
zhanglongxia: Have updated.

zhanglongxia:
@abtink  Could you review this PR at your convenience?

abtink:
Looks great. Thanks. :+1: 

One suggestion: Now that we have the capability, how adding the `"NCP:CoexMetrics"` and `"NCP:Coex:Enable"` to all supported property list in [`SpinelNCPInstance::get_supported_property_keys()`](https://github.com/openthread/wpantund/blob/master/src/ncp-spinel/SpinelNCPInstance.cpp#L442)? We can then see them in the full list from `wpanctl get`.

PR #437, opened 2019-10-01
[style] fix conversion warnings in string-utils

bukepo:


bukepo:
@abtink any comments?

PR #438, opened 2019-10-03
Feature to auto add on-mesh prefixes as interface route

abtink:
This commit adds a new feature in `wpantund` to auto add a route on
the host primary interface corresponding to on-mesh prefixes from
Thread network.

This feature can be enabled through a newly added wpan property with
name "Daemon:OnMeshPrefix:AutoAddAsIntterfaceRoute" which is enabled
by default.

When enabled, wpantund would add a route on host primary interface for
any prefix from thread network (with on-mesh flag set). This in turn
ensures that traffic destined to an IPv6 address matching the prefix
would be correctly forwarded to the `wpan` interface on host.

----
This feature is now independent of `mAutoAddOffMeshRoutesOnInterface` (each feature can be separately enabled or disabled).

https://github.com/openthread/openthread/pull/4221 adds a test-case to cover the behavior of this feature.

### DIFF
--- a/src/wpantund/NCPInstanceBase-Addresses.cpp
+++ b/src/wpantund/NCPInstanceBase-Addresses.cpp
@@ -929,6 +929,11 @@ NCPInstanceBase::on_mesh_prefix_was_added(Origin origin, const struct in6_addr &
 		} else {
 			cb(kWPANTUNDStatus_Ok);
 		}
+
+		if (mAutoAddOnMeshPrefixesAsInterfaceRoutes) {
+			mRequestRouteRefresh = true;
+		}
+
 	} else {
 		cb(kWPANTUNDStatus_Ok);
 	}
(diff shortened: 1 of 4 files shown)


src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -1207,6 +1231,24 @@ NCPInstanceBase::refresh_routes_on_interface(void)
 		}
 	}
 
+	if (mAutoAddOnMeshPrefixesAsInterfaceRoutes) {
+		// Iterate through all off-mesh route entries to check if there is a new route that should be added on interface.
```
jwhui: nit: "off-mesh route" -> "on-mesh prefix"?

jwhui:
Should we consider the use case of auto adding routes for on-mesh prefixes but not for external routes? This would cover the scenario that a device only cares to forward traffic destined for Thread devices.

abtink:
Good point! Will change it to have them be separate features.

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -1076,68 +1085,88 @@ bool
 NCPInstanceBase::should_add_route_on_interface(const IPv6Prefix &route, uint32_t &metric)
 {
 	bool should_add = false;
-	bool route_added_by_device = false;
-	bool route_added_by_others = false;
-	RoutePreference preference_device = NCPControlInterface::ROUTE_LOW_PREFRENCE;
-	RoutePreference preference_others = NCPControlInterface::ROUTE_LOW_PREFRENCE;
 
-	std::multimap<IPv6Prefix, OffMeshRouteEntry>::iterator iter, sub_iter, upper_iter;
+	if (mAutoAddOffMeshRoutesOnInterface) {
```
abtink: This is just moving all the off-mesh route related code under a `if (mAutoAddOffMeshRoutesOnInterface)` check.
Diff does look much simpler and more readable with "hide white-space changes".

PR #440, opened 2019-10-18
Fix compile warnings

zhanglongxia:


src/util/socket-utils.c
```diff
@@ -352,6 +352,10 @@ lookup_sockaddr_from_host_and_port(
 }
 
 #if HAVE_FORKPTY
+
+// Decalre function to avoid compile warning: implicit declaration of function ‘posix_openpt’
```
abtink: nit - typo: `Decalre` -> `Declare`

zhanglongxia: Updated.

PR #444, opened 2019-11-08
Listen for MLD report to add/remove multicast groups

gjc13:
Currently using IP6_JOIN_GROUP to listen on multicast addresses won't
notify wpantund. This patch listens for a MLD report and updates the
multicast group to the NCP.

Test script:

``` python
#!/usr/bin/env python3

import socket
import struct
import sys

s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)

s.bind(('ff02::158', 30000))
s.setsockopt(socket.IPPROTO_IPV6,       socket.IPV6_MULTICAST_IF, int(sys.argv[1]))
s.setsockopt(socket.IPPROTO_IPV6,       socket.IPV6_JOIN_GROUP, 
    struct.pack('16si', socket.inet_pton(socket.AF_INET6, 'ff02::158'), int(sys.argv[1])))

while True:
  print(s.recvfrom(4096))
```

Run wpantund and run the script with the first argument as the wpan0 interface index.
You can see the multicast group added to and removed from the NCP when the script starts/exits.

### DIFF
--- a/src/util/TunnelIPv6Interface.cpp
+++ b/src/util/TunnelIPv6Interface.cpp
@@ -37,6 +37,7 @@
 #include <linux/netlink.h>
 #include <linux/rtnetlink.h>
 #include <fcntl.h>
+#include <ifaddrs.h>
 #endif
 
 #include <sys/select.h>
@@ -44,6 +45,43 @@
 #ifndef O_NONBLOCK
 #define O_NONBLOCK          O_NDELAY
 #endif
+
+namespace {
+
+	const char* kMLDv2MulticastAddress= "ff02::16";
+
+	const char *kFilterMulticastAddresses[] = {
+			kMLDv2MulticastAddress,
+			"ff02::01", //Link local all nodes
+			"ff02::02", //Link local all routers
+			"ff03::01", //realm local all nodes
+			"ff03::02", //realm local all routers
+			"ff03::fc", //realm local all mpl
+	};
+
+	struct MLDv2Header {
+		uint8_t mType;
+		uint8_t _rsv0;
+		uint16_t mChecksum;
+		uint16_t _rsv1;
+		uint16_t mNumRecords;
+	} __attribute__((packed));
+
+	struct MLDv2Record {
+		uint8_t mRecordType;
+		uint8_t mAuxDataLen;
+		uint16_t mNumSources;
+		struct in6_addr mMulticastAddress;
+		struct in6_addr mSourceAddresses[];
+	} __attribute__((packed));
+
+} // namespace
+
+static bool
+is_addr_multicast(const struct in6_addr &address)
+{
+	return (address.s6_addr[0] == 0xff);
+}
 
 TunnelIPv6Interface::TunnelIPv6Interface(const std::string& interface_name, int mtu):
 #if FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION
@@ -59,6 +97,7 @@ TunnelIPv6Interface::TunnelIPv6Interface(const std::string& interface_name, int 
 #else
 	mNetifMgmtFD(netif_mgmt_open()),
 #endif
+	mMLDMonitorFD(-1),
 	mIsRunning(false),
 	mIsUp(false)
 {
@@ -88,12 +127,16 @@ TunnelIPv6Interface::TunnelIPv6Interface(const std::string& interface_name, int 
 	netif_mgmt_set_mtu(mNetifMgmtFD, mInterfaceName.c_str(), mtu);
 
 	setup_signals();
+	setup_mld_listener();
 #endif
 }
 
 TunnelIPv6Interface::~TunnelIPv6Interface()
 {
 	close(mNetlinkFD);
+	if (mMLDMonitorFD >= 0) {
+		close(mMLDMonitorFD);
+	}
 	netif_mgmt_close(mNetifMgmtFD);
 }
 
@@ -119,7 +162,7 @@ TunnelIPv6Interface::on_link_state_changed(bool isUp, bool isRunning)
 				iter->second.mState = Entry::kWaitingForAddConfirm;
 			}
 
-			for (iter = mMulticastAddresses.begin(); iter != mMulticastAddresses.end();	++iter) {
+			for (iter = mPendingMulticastAddresses.begin(); iter != mPendingMulticastAddresses.end();	++iter) {
 				if (iter->second.mState != Entry::kWaitingToAdd) {
 					continue;
 				}
@@ -130,7 +173,7 @@ TunnelIPv6Interface::on_link_state_changed(bool isUp, bool isRunning)
 				IGNORE_RETURN_VALUE(netif_mgmt_join_ipv6_multicast_address(mNetifMgmtFD, mInterfaceName.c_str(),
 				                    iter->first.s6_addr));
 			}
-			mMulticastAddresses.clear();
+			mPendingMulticastAddresses.clear();
 		}
 		mIsUp = isUp;
 		mIsRunning = isRunning;
@@ -148,7 +191,20 @@ TunnelIPv6Interface::on_address_added(const struct in6_addr &address, uint8_t pr
 	syslog(LOG_INFO, "TunnelIPv6Interface: \"%s/%d\" was added to \"%s\"", in6_addr_to_string(address).c_str(), prefix_len,
 	       get_interface_name().c_str());
 
-	mAddressWasAdded(address, prefix_len);
+	mUnicastAddressWasAdded(address, prefix_len);
+}
+
+void
+TunnelIPv6Interface::on_multicast_address_joined(const struct in6_addr &address)
+{
+	if (mPendingMulticastAddresses.count(address)) {
+		mPendingMulticastAddresses.erase(address);
+	}
+
+	syslog(LOG_INFO, "TunnelIPv6Interface: \"%s\" was added to \"%s\"", in6_addr_to_string(address).c_str(),
+	       get_interface_name().c_str());
+
+	mMulticastAddressWasJoined(address);
 }
 
 void
@@ -164,11 +220,27 @@ TunnelIPv6Interface::on_address_removed(const struct in6_addr &address, uint8_t 
 		syslog(LOG_INFO, "TunnelIPv6Interface: \"%s/%d\" was removed from \"%s\"", in6_addr_to_string(address).c_str(), prefix_len,
 		       get_interface_name().c_str());
 
-		mAddressWasRemoved(address, prefix_len);
+		mUnicastAddressWasRemoved(address, prefix_len);
 	}
 
 }
 
+void
+TunnelIPv6Interface::on_multicast_address_left(const struct in6_addr &address)
+{
+	// Ignore "removed" signal if address is the list
+	// meaning it was added earlier and we are still
+	// waiting to confirm that it is added.
+	// This is to address the case where before adding an
+	// address on interface it may be first removed.
+
+	if (!mPendingMulticastAddresses.count(address)) {
+		syslog(LOG_INFO, "TunnelIPv6Interface: \"%s\" was removed from \"%s\"", in6_addr_to_string(address).c_str(),
+		       get_interface_name().c_str());
+
+		mMulticastAddressWasLeft(address);
+	}
+}
 
 #if __linux__ // --------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/util/TunnelIPv6Interface.h
```diff
@@ -79,25 +79,36 @@ class TunnelIPv6Interface : public nl::UnixSocket
...
+	void processNetlinkFD(void);
+	void processMLDMonitorFD(void);
 
 	void on_link_state_changed(bool isUp, bool isRunning);
 	void on_address_added(const struct in6_addr &address, uint8_t prefix_len);
+	void on_multicast_address_joined(const struct in6_addr &address);
 	void on_address_removed(const struct in6_addr &address, uint8_t prefix_len);
+	void on_multicast_address_left(const struct in6_addr &address);
+
+	bool is_addr_multicast(const struct in6_addr &address);
```
bukepo: should be a static or local function.

src/util/TunnelIPv6Interface.h
```diff
@@ -114,7 +125,31 @@ class TunnelIPv6Interface : public nl::UnixSocket
 			mState(state) { }
 	};
 
+	struct MLDv2Header {
```
bukepo: This is only used in implementation file, move this into the .cpp file?

src/util/TunnelIPv6Interface.h
```diff
@@ -114,7 +125,31 @@ class TunnelIPv6Interface : public nl::UnixSocket
...
+		uint8_t mType;
+		uint8_t _rsv0;
+		uint16_t mChecksum;
+		uint16_t _rsv1;
+		uint16_t mNumRecords;
+	} __attribute__((packed));
+
+	struct MLDv2Record {
```
bukepo: move into .cpp file?

src/util/TunnelIPv6Interface.h
```diff
@@ -114,7 +125,31 @@ class TunnelIPv6Interface : public nl::UnixSocket
...
+		struct in6_addr mMulticastAddress;
+		struct in6_addr mSourceAddresses[];
+	} __attribute__((packed));
+
 	std::map<struct in6_addr, Entry> mUnicastAddresses;
-	std::map<struct in6_addr, Entry> mMulticastAddresses;
+	std::map<struct in6_addr, Entry> mPendingMulticastAddresses;
+	std::set<struct in6_addr> mMulticastAddresses;
+
+	static const char* kMLDv2MulticastAddress;
```
bukepo: declare as a file scope variable in the .cpp file?

gjc13:
Now we are having two interfaces to join a multicast group, one via setsockopt and the other via wpanctl. @abtink Shall we mark the old wpanctl manner as deprecated?

src/util/TunnelIPv6Interface.cpp
```diff
@@ -214,8 +263,43 @@ TunnelIPv6Interface::setup_signals(void)
...
+	mreq6.ipv6mr_interface = interfaceIndex;
+	inet_pton(AF_INET6, kMLDv2MulticastAddress, &mreq6.ipv6mr_multiaddr);
+
+	require(setsockopt(mMLDMonitorFD, IPPROTO_IPV6, IPV6_JOIN_GROUP, &mreq6, sizeof(mreq6)) == 0, bail);
+	require(setsockopt(mMLDMonitorFD, SOL_SOCKET, SO_BINDTODEVICE, mInterfaceName.c_str(), mInterfaceName.size()) == 0, bail);
+
+	success = true;
+
+bail:
+	if (!success) {
```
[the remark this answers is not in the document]
gjc13: Failing to listen to MLD packets doesn't prevent other functionalities from running so I chose to output an error message.

src/util/TunnelIPv6Interface.cpp
```diff
@@ -148,7 +182,24 @@ TunnelIPv6Interface::on_address_added(const struct in6_addr &address, uint8_t pr
...
+		mPendingMulticastAddresses.erase(address);
+	}
+
+	syslog(LOG_INFO, "TunnelIPv6Interface: \"%s\" was added to \"%s\"", in6_addr_to_string(address).c_str(),
+	       get_interface_name().c_str());
+
+	if (mMulticastAddresses.count(address) == 0) {
+		// update new addresses only
+		mMulticastAddresses.insert(address);
+		mMulticastAddressWasJoined(address);
```
abtink: I think for unicast address we emit the signal on any address add (we don't maintain a list of unicast address in this class and check if it's already in list or not from this class before emitting `mUnicastAddressWasAdded`). 

Can we consider doing the same for multicast addresses? 
If we go for this, I think we won't need the `mMulticastAddresses`  (we just need a pending list of multicast addresses to add which would be simpler).

Note that the `NCPInstanceBase` class itself will manage all address types (and handle the `mMulticastAddressWasJoined` signal. It maintains list of multicast addresses and it would handle if address is already in the list. 

Thoughts?

abtink:
It's great that you posted a test script to verify the added feature and its behavior.

Would it be possible (perhaps from a separate future  PR) to add such a test-case to verify addition/removal of multicast addresses (using new model) under Travis? I think we may able to update `toranj` to add such a test-case.

Thoughts?

src/util/TunnelIPv6Interface.cpp
gjc13: The problem I have intended to solve here is that because we are joining all the multicast groups the NCP has joined by itself (at line 663 in the same file). This will also trigger a MLD packet which will also be captured by the listener. The `mMulticastAddresses` has been added to prevent joining the group again.

gjc13:
> It's great that you posted a test script to verify the added feature and its behavior.
> 
> Would it be possible (perhaps from a separate future PR) to add such a test-case to verify addition/removal of multicast addresses (using new model) under Travis? I think we may able to update `toranj` to add such a test-case.
> 
> Thoughts?

Currently we are using the openthread test framework but I'd be eager to add another test script in a separate PR.

src/util/TunnelIPv6Interface.cpp
abtink: As you say, OpenThread itself already joins the same multicast address and `wpantund` would sync addresses between OT/NCP and host interface. I think the addition of the same address in `wpantund` (line 663) is actually redundant and can be possibly removed. This would then simplify the code (remove the need for `mMulticastAddresses` in this class).

Thoughts?

gjc13: The addition of the same address at line 663 is required to notify the linux kernel to not filter these addresses.

abtink: Let me first clarify my earlier comment. I realized we may have been talking about two different things :)...I initially thought your were referring to this line (where from constructor we join a multicast group eventhough OT/NCP itself would join the same address):

https://github.com/openthread/wpantund/blob/master/src/wpantund/NCPInstanceBase.cpp#L149-L151

This is what I refereed to as redundant in my earlier comment - not line 663.
I think we can maybe address this later/different PR.

---

Regarding multicast addresses originating from NCP and added on linux interface, I still think we don't need a separate `mMulticastAddresses` in `TunnelIPv6Interface` class.

The `NCPInstanceBase` class already keeps track of all multicast addresses and also  tracks their origin (who added the address, NCP, interface, user). The `TunnelIPv6Interface`  does not need to track/save list of addresses, it just needs to provide API to add/remove addresses and signals to notify when an address is added/removed (from interface). The `NCPInstanceBase` would not re-add an address again if it's already in its list. 
- NCP reports a new multicast address
- `NCPInstanceBase` would add it in its list (marking origin as NCP) and then request it to be added on linux interface (calls `TunnelIPv6Interface::join_multicast_address()`.
- A bit later (when address is actually added on interface), we get it in `TunnelIPv6Interface`  through MLD. 
- At this point, it is ok for `TunnelIPv6Interface` to invoke `mMulticastAddressWasJoined` signal (which is hooked to `NCPInstanceBase::multicast_address_was_joined()`).
- The `mlticast_address_was_joined()` would first check if the address is new or not (if it's already in its list) and takes no action.
https://github.com/openthread/wpantund/blob/master/src/wpantund/NCPInstanceBase-Addresses.cpp#L767-L792

This is same pattern we use for unicast addresses already. My suggestion was to do the same for multi-cast (unless there is clear difference between unicast/multicast...) If we can go for this we can `mMulticastAddresses` in `TunnelIPv6Interface` and make it simpler.

Thoughts?

gjc13: I've tried to remove mMulticastAddresses but it caused the OpenThread toranj test to fail. I'll investigate what's going wrong.

abtink: Can you please add which tests are failing (with removal of `mMulticastAddresses`)?

gjc13: It's test-02 test-join which reports that the device is not in offline state when `wpanctl leave` is called. Waiting for about a second suffices to put the device to the right state.

gjc13: On my workstation it's test-002-form but it works fine on the travis ci. I'll double check what's going wrong.

gjc13:
It turned out that when we are leaving a Thread network, the kernel will send the full list of IPv6 multicast addresses it's listening to as well as the addresses it's leaving in the MLD listener update. On my workstation, for one reason or another, it's always listening to ff02::2. We tried to add the address to the NCP when it is resetting itself, causing a NCP misbehaving event and the NCP will be reset again. 

Such case can happen when there's really a user listening to a multicast address on the wpan interface. Under such circumstance, the kernel will also include this address in the MLD report as well and this address will be added to the NCP in an invalid state.

A simple fix is to check whether the NCP is ready (not in uninitialized state) when adding an address but it will cause the user to silently lose the access to the multicast group it has joined. Will this be a severe issue?

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -775,8 +775,7 @@ NCPInstanceBase::multicast_address_was_joined(Origin origin, const struct in6_ad
 		if ((origin == kOriginThreadNCP) || (origin == kOriginUser)) {
 			mPrimaryInterface->join_multicast_address(&address);
 		}
-
-		if (origin == kOriginPrimaryInterface) {
+		if (origin == kOriginPrimaryInterface && !ncp_state_is_initializing(get_ncp_state())) {
```
abtink: Not sure about not allowing addresses to be updated while NCP is being initalized.
With this we can potentially miss adding a unicast/multicast address added on interface (causing a miss-match between interface and NCP).

abtink:
@gjc13 , I think we need to understand why it fails when we add a multicast address during NCP init? Technically we should be able to update addresses on NCP/OpenThread at any point.

abtink:
@gjc13 just submitted https://github.com/openthread/openthread/pull/4388 which now allows multicast addresses to be added/removed at any state in NCP.
But in new code, we still do not allow users to add certain fixed addresses (these are managed by OT `Netif` itself).

I think we need to update wpantund to ensure we ignore these addresses if added by user or through interface:
- ff02::01    LinkLocalAllNodes   
- ff03::01    RealmLocalAllNodes  
- ff03::fc    RealmLocalAllMpl    
- ff02::02    LinkLocalAllRouters 
- ff03::02    RealmLocalAllRouters

This would be similar to how we already handle user/interface-added link-local unicast addresses. Within OT/NCP we do not allow users to add a link-local unicast address, so in wpantund we have code to filter them https://github.com/openthread/wpantund/blob/87c90eedce0c75cb68a1cbc34ff36223400862f1/src/wpantund/NCPInstanceBase-Addresses.cpp#L635-L646
I think we need similar logic for multicast address.

Thoughts?

gjc13:
> @gjc13 just submitted [openthread/openthread#4388](https://github.com/openthread/openthread/pull/4388) which now allows multicast addresses to be added/removed at any state in NCP.
> But in new code, we still do not allow users to add certain fixed addresses (these are managed by OT `Netif` itself).
> 
> I think we need to update wpantund to ensure we ignore these addresses if added by user or through interface:
> 
> * ff02::01    LinkLocalAllNodes
> * ff03::01    RealmLocalAllNodes
> * ff03::fc    RealmLocalAllMpl
> * ff02::02    LinkLocalAllRouters
> * ff03::02    RealmLocalAllRouters
> 
> This would be similar to how we already handle user/interface-added link-local unicast addresses. Within OT/NCP we do not allow users to add a link-local unicast address, so in wpantund we have code to filter them
> 
> https://github.com/openthread/wpantund/blob/87c90eedce0c75cb68a1cbc34ff36223400862f1/src/wpantund/NCPInstanceBase-Addresses.cpp#L635-L646
> 
> 
> I think we need similar logic for multicast address.
> Thoughts?

A filter has been added in the latest patch.

PR #452, opened 2020-01-01
Handle timestamp log metadata received from NCP

konradderda:
* Handle timestamp field in log metadata introduced by:
https://github.com/openthread/openthread/pull/4439

* Add new property "OpenThread:LogTimestampBase" that can be used to change the NCP log base timestamp

This change does not break the compatibility: older wpantund will work with newer OpenThread and newer wpantund will work with older OpenThread.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance.cpp
@@ -232,7 +232,7 @@ SpinelNCPInstance::handle_ncp_log_stream(const uint8_t *data_in, int data_len)
 		data_in += len;
 		data_len -= len;
 
-		if(data_len >= sizeof(log_timestamp)) {
+		if (data_len >= sizeof(log_timestamp)) {
 			len = spinel_datatype_unpack(
 				data_in,
 				data_len,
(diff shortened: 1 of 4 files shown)


third_party/openthread/src/ncp/spinel.h
```diff
@@ -3896,6 +3896,14 @@ typedef enum
      */
     SPINEL_PROP_DEBUG_TEST_WATCHDOG = SPINEL_PROP_DEBUG__BEGIN + 2,
 
+    /// The NCP timestamp base
+    /** Format: X (write-only) 
+     * 
+     * This property controls the time base value that is used for log timestamp field calulation.
```
LuDuda: nit: typo calculation

src/ncp-spinel/SpinelNCPInstance.cpp
```diff
@@ -215,24 +215,49 @@ SpinelNCPInstance::handle_ncp_log_stream(const uint8_t *data_in, int data_len)
...
-			prefix_string,
-			sizeof(prefix_string),
-			"[%s]%s: ",
-			ot_log_level_to_string(log_level),
-			ot_log_region_to_string(log_region)
-		);
+		data_in += len;
+		data_len -= len;
+
+		if(data_len >= sizeof(log_timestamp)) {
```
LuDuda: nit: add extra space after if

PR #454, opened 2020-01-21
[util] remove strlcpy reference

gjc13:
strlcpy is sometimes only found in libbsd. Use a hand-written implementation instead to avoid link errors.

### DIFF
--- a/src/util/netif-mgmt.c
+++ b/src/util/netif-mgmt.c
@@ -73,6 +73,23 @@
 #define SIOCSIFLLADDR SIOCSIFHWADDR
 #endif
 
+// incase strlcpy is not found in libc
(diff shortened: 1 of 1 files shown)


src/util/netif-mgmt.c
```diff
@@ -73,6 +73,23 @@
 #define SIOCSIFLLADDR SIOCSIFHWADDR
 #endif
 
+// incase strlcpy is not found in libc
+size_t _my_strlcpy(char *aDst, const char *aSrc, size_t aSize)
```
abtink: Not sure if we would need this since `wpantund` provides/uses its own `strlcpy` (if not available) from `src/missing/strlcpy` file.
Check this source https://github.com/openthread/wpantund/blob/master/src/missing/strlcpy/strlcpy.c#L30

PR #455, opened 2020-01-23
[util] include missing strlcpy header

gjc13:
Always use the missing strlcpy header in case strlcpy is not provided
in the system library. This header will define strlcpy as a macro which
points to __missing_strlcpy.


bukepo:
Update the Android.mk as well?

PR #457, opened 2020-01-31
[build] support autoconf-archive 2019.01.06

bukepo:
This PR adds support for autoconf-archive 2019.01.06, which doesn't
define CODE_COVERAGE_RULES anymore.


bukepo:
@abtink any comments?

PR #458, opened 2020-02-04
[build] fix bootstrap for autoconf-archive 2019.01.16

bukepo:
`-include` doesn't work as expected on Makefile.am. This PR fixes it by always generating  `aminclude_static.am`.


abtink:
@bukepo  thanks for helping with this and very quick fix. :+1: :+1:

PR #472, opened 2020-06-24
Add support for Discerner for Joiner/Commissioner

abtink:
This commit adds support for Joiner Discerner in `wpantund` and `wpanctl`.  The Joiner Discerner enables new a mechanism for Thread commissioning. The traditional Thread commissioning process uses factory assigned EUI-64 of the device to derive the Joiner ID and identify/filter a joiner (through steering data bloom filter). The Joiner Discerner (which is an unsigned value along with a user-specified bit length up to 64 bits) allows users to have more control and do not rely on factory-assigned EUI-64.

For support of Discerner for Joiner operation in `wpantund` two methods are provided: Support for two new wpan properties `Joiner:Discerner:Value` and `Joiner:Discerner:BitLength` are added. To set the Joiner Discerner on device, the bit-length must be set first followed by setting the value. Setting the bit-length to zero will clear any previously set Joiner Discerner on the NCP (i.e., device would go back to default behavior using the factory-assigned EUI-64). Alternatively, the discerner value can be specified as the value-map parameters (with same keys as wpan property names) passed along through a "Joiner Start" DBus API call.

For commissioner, the NCP Control Interface related APIs and the corresponding DBus APIs are updated to allow either an EUI-64 or a Joiner Discerner (or accepting any Joiner) to be provided. `wpanctl` `commissioner` command is also updated to allow the same.

-----
Related PR in openthread: https://github.com/openthread/openthread/pull/5137

### DIFF
--- a/src/wpanctl/tool-cmd-commr.c
+++ b/src/wpanctl/tool-cmd-commr.c
@@ -484,7 +484,7 @@ tool_cmd_commr(int argc, char* argv[])
 		uint8_t discerner_bit_len = 0;
 		uint32_t joiner_timeout = 0;
 
-		// joiner-remove-with-discerner <discerner value> <discerner bit length> [<timeout>=0]
+		// joiner-remove-discerner <discerner value> <discerner bit length> [<timeout>=0]
 
 		memset(&eui64, 0, sizeof(eui64));
 
--- a/src/wpanctl/tool-cmd-commr.c
+++ b/src/wpanctl/tool-cmd-commr.c
@@ -385,7 +385,7 @@ tool_cmd_commr(int argc, char* argv[])
 		uint32_t joiner_timeout;
 		const char *psk;
 
-		// joiner-add-with-discerner <discerner value> <discerner bit length> <timeout> <psk>
+		// joiner-add-discerner <discerner value> <discerner bit length> <timeout> <psk>
 
 		memset(&eui64, 0, sizeof(eui64));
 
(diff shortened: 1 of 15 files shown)


src/wpanctl/tool-cmd-commr.c
```diff
@@ -361,6 +376,52 @@ tool_cmd_commr(int argc, char* argv[])
...
+	case COMMR_CMD_JOINER_ADD_WITH_DISCERNER:
+	{
+		uint8_t eui64[COMMR_EUI64_SIZE];
+		uint8_t *eui64_ptr = eui64;
+		uint64_t discerner_value = 0;
+		uint8_t discerner_bit_len = 0;
+		uint32_t joiner_timeout;
+		const char *psk;
+
+		// joiner-add-with-discerner <discerner value> <discerner bit length> <timeout> <psk>
```
jwhui, line 388:
```suggestion
		// joiner-add-discerner <discerner value> <discerner bit length> <timeout> <psk>
```

src/wpanctl/tool-cmd-commr.c
```diff
@@ -415,6 +476,43 @@ tool_cmd_commr(int argc, char* argv[])
...
 
+	case COMMR_CMD_JOINER_REMOVE_WITH_DISCERNER:
+	{
+		uint8_t eui64[COMMR_EUI64_SIZE];
+		uint8_t *eui64_ptr = eui64;
+		uint64_t discerner_value = 0;
+		uint8_t discerner_bit_len = 0;
+		uint32_t joiner_timeout = 0;
+
+		// joiner-remove-with-discerner <discerner value> <discerner bit length> [<timeout>=0]
```
jwhui, line 487:
```suggestion
		// joiner-remove-discerner <discerner value> <discerner bit length> [<timeout>=0]
```

PR #482, opened 2021-02-15
[spinel] update domain prefix spinel property

XPAWAN:
This PR update the wpantund domain prefix command with new Nd Dns and Domain Prefix flag.
The SPINEL_PROP_THREAD_ON_MESH_NETS property has been extended with new TLV flags field.

### DIFF
--- a/src/ipc-dbus/DBusIPCAPI_v1.cpp
+++ b/src/ipc-dbus/DBusIPCAPI_v1.cpp
@@ -1241,6 +1241,8 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
 	dbus_bool_t dhcp = FALSE;
 	dbus_bool_t configure = FALSE;
 	dbus_bool_t stable = TRUE;
+	dbus_bool_t nd_dns = FALSE;
+	dbus_bool_t domain_prefix = FALSE;
 	uint32_t preferred_lifetime = 0;
 	uint32_t valid_lifetime = 0;
 	uint8_t *prefix(NULL);
@@ -1267,8 +1269,29 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
 		DBUS_TYPE_BOOLEAN, &configure,
 		DBUS_TYPE_BOOLEAN, &stable,
 		DBUS_TYPE_UINT16, &prefix_len_in_bits,
+		DBUS_TYPE_BOOLEAN, &nd_dns,
+		DBUS_TYPE_BOOLEAN, &domain_prefix,
 		DBUS_TYPE_INVALID
 	);
+
+	if (!did_succeed) {
+		did_succeed = dbus_message_get_args(
+			message, NULL,
+			DBUS_TYPE_BOOLEAN, &default_route,
+			DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &prefix, &prefix_len_in_bytes,
+			DBUS_TYPE_UINT32, &preferred_lifetime,
+			DBUS_TYPE_UINT32, &valid_lifetime,
+			DBUS_TYPE_BOOLEAN, &preferred,
+			DBUS_TYPE_BOOLEAN, &slaac,
+			DBUS_TYPE_BOOLEAN, &on_mesh,
+			DBUS_TYPE_INT16, &priority_raw,
+			DBUS_TYPE_BOOLEAN, &dhcp,
+			DBUS_TYPE_BOOLEAN, &configure,
+			DBUS_TYPE_BOOLEAN, &stable,
+			DBUS_TYPE_UINT16, &prefix_len_in_bits,
+			DBUS_TYPE_INVALID
+		);
+	}
 
 	if (!did_succeed) {
 		did_succeed = dbus_message_get_args(
@@ -1346,6 +1369,14 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
 
 	if (configure) {
 		flags.insert(NCPControlInterface::PREFIX_FLAG_CONFIGURE);
+	}
+
+	if (nd_dns) {
+		flags.insert(NCPControlInterface::PREFIX_FLAG_ND_DNS);
+	}
+
+	if (domain_prefix) {
+		flags.insert(NCPControlInterface::PREFIX_FLAG_DOMAIN_PREFIX);
 	}
 
 	dbus_message_ref(message);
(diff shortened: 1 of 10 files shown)


XPAWAN:
Related PR into OpenThread: openthread/openthread#6174

abtink (approved):
Looks really great. Thanks 👍

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -175,14 +175,16 @@ NCPInstanceBase::on_mesh_prefix_flags_to_string(uint8_t flags, bool detailed)
 			break;
 		}
 
-		snprintf(c_string, sizeof(c_string), "flags:0x%02x [on-mesh:%s def-route:%s config:%s dhcp:%s slaac:%s pref:%s prio:%s]",
+		snprintf(c_string, sizeof(c_string), "flags:0x%02x [on-mesh:%s def-route:%s config:%s dhcp:%s slaac:%s pref:%s nd-dns:%s dp:%s prio:%s]",
```
abtink, line 178: @XPAWAN I noticed that this change is not passing CI test.

Seems like it's due to this change here which impacts the output of `wpanctl` command which is used by `toranj/wpan.py` to parse and get the flags:
https://github.com/openthread/openthread/blob/f57c5e23709c33d6b0c1a4e209223f35fd040529/tests/toranj/wpan.py#L1176-L1204

------

I don't see any way to update this in `wpantund` without breaking the compatibility with current `wpan.py` implementation.

Here is what I can think of and suggest as way to move forward:
- In a PR in OT repo, we update the `wpan.py` to allow extra chars in between `pref:` and `prio:` (and just skips over it without parsing the content)
- We would then need to merge that first and the change here from this PR in `wpantund` should pass and we can merge this PR.
- After that we can later consider updating the `wpan.py` to actually parse the new flags.

Any other ideas?

abtink: @XPAWAN I submitted https://github.com/openthread/openthread/pull/6240 which relaxes the parsing of the prefix list `wpanctl` output in `troanj` (what I mentioned above).

However, I noticed even with this change the test were still failing (on removing of prefix) -> https://github.com/abtink/openthread/runs/2043151764?check_suite_focus=true

I think there is some other missing piece here in this PR in `wpantund` itself.
- This PR is changing the DBUS commenad (adding new flags to it).
- The `tool-cmd-add-prefix.c` is updated to use the new DBUS message format, but the `tool-cmd-remove-prefix` is still using old DBUS format (without new flags). 
- On the `wpantund` side the PR is parsing DBUs message expecting the new flags to be present and so the `remove-prefix` is failing

src/ipc-dbus/DBusIPCAPI_v1.cpp
```diff
@@ -1267,6 +1269,8 @@ DBusIPCAPI_v1::interface_config_gateway_handler(
 		DBUS_TYPE_BOOLEAN, &configure,
 		DBUS_TYPE_BOOLEAN, &stable,
 		DBUS_TYPE_UINT16, &prefix_len_in_bits,
+		DBUS_TYPE_BOOLEAN, &nd_dns,
+		DBUS_TYPE_BOOLEAN, &domain_prefix,
 		DBUS_TYPE_INVALID
 	);
```
abtink, line 1276: @XPAWAN , follow up from https://github.com/openthread/wpantund/pull/482#discussion_r588782383, we have two ways to fix 
1) Update the `tool-cmd-remove-prefix` to also add the `nd_ds` and `domain_prefix` flags in DBus message it sends
2) Make sure the parser here handle both cases (with/without extra flags)

The second option is much better, since it keeps the DBus API of wpantund backward-compatible.

We generally aim to keep the DBus API backward compatible (unless there is a good justification/reason for it). This is to ensure newer versions of `wpantund` continues to work with older modules/processes that may interact with `wpantund` through its DBus API.

Here is my suggestion to add this here:

```
	if (!did_succeed) {
		did_succeed = dbus_message_get_args(
			message, NULL,
			DBUS_TYPE_BOOLEAN, &default_route,
			DBUS_TYPE_ARRAY, DBUS_TYPE_BYTE, &prefix, &prefix_len_in_bytes,
			DBUS_TYPE_UINT32, &preferred_lifetime,
			DBUS_TYPE_UINT32, &valid_lifetime,
			DBUS_TYPE_BOOLEAN, &preferred,
			DBUS_TYPE_BOOLEAN, &slaac,
			DBUS_TYPE_BOOLEAN, &on_mesh,
			DBUS_TYPE_INT16, &priority_raw,
			DBUS_TYPE_BOOLEAN, &dhcp,
			DBUS_TYPE_BOOLEAN, &configure,
			DBUS_TYPE_BOOLEAN, &stable,
			DBUS_TYPE_UINT16, &prefix_len_in_bits,
			DBUS_TYPE_INVALID
		);
	}
```

I think/expect with such a change here and https://github.com/openthread/openthread/pull/6240 in OT, the CI tests should all pass.

abtink (changes requested):

src/ipc-dbus/DBusIPCAPI_v1.cpp
XPAWAN: Hello @abtink. Thank You for Your support. 
I have added a new commit according to Your remarks.
The OT related PR is already merged( openthread/openthread#6174 )
Shall I also create a new official PR with changes for openthread as in  https://github.com/openthread/openthread/pull/6240 ?

abtink: I just re-ran the test and it did indeed pass ok 
- https://github.com/openthread/openthread/pull/6240/checks?check_run_id=2058442009

--------

> Shall I also create a new official PR with changes for openthread as in openthread/openthread#6240 ?

I will submit it as PR now (removing the top commit with fixed it to use the `wpantund` from this branch).

Here is the plan:
- We merge the https://github.com/openthread/openthread/pull/6240 first.
- After that, we can re-run  CI on this PR (which should hopefully then pass).
- Then we can then merge this PR.

abtink:
@XPAWAN can you please help rebase this branch on master? We just merged https://github.com/openthread/wpantund/pull/487 which fixes/re-enables CI to use the recently renamed OT repro `main` branch name and we need to it to re-run the CI tests. Thanks.

abtink (approved):
👍

PR #483, opened 2021-02-24
[spinel] add interface identifier of the Thread Domain Unicast Address

XPAWAN:
This commit add a  new spinel property to configure Interface Identifier of the Thread Domain Unicast Address.
OT PR: https://github.com/openthread/openthread/pull/6200


abtink (approved):
LGTM. 👍

PR #489, opened 2021-03-19
Ensure only one Domain Prefix entry in on mesh prefixes

XPAWAN:


src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -918,38 +918,80 @@ NCPInstanceBase::find_prefix_entry(const IPv6Prefix &prefix, const OnMeshPrefixE
 	return iter;
 }
 
+std::multimap<NCPInstanceBase::IPv6Prefix, NCPInstanceBase::OnMeshPrefixEntry>::iterator
+NCPInstanceBase::find_domain_prefix_entry(const IPv6Prefix &prefix, const OnMeshPrefixEntry &entry)
```
abtink, line 922: We don't seem to use `entry`. Can we consider removing it from list of parameters?

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -918,38 +918,80 @@ NCPInstanceBase::find_prefix_entry(const IPv6Prefix &prefix, const OnMeshPrefixE
...
+
+
 void
 NCPInstanceBase::on_mesh_prefix_was_added(Origin origin, const struct in6_addr &prefix_address, uint8_t prefix_len,
 	uint16_t flags, bool stable, uint16_t rloc16, CallbackWithStatus cb)
 {
 	IPv6Prefix prefix(prefix_address, prefix_len);
 	OnMeshPrefixEntry entry(origin, flags, stable, rloc16);
 	std::multimap<IPv6Prefix, OnMeshPrefixEntry>::iterator iter;
+	bool add_new_domain_entry = false;
```
abtink, line 947: Can we rename this as `add_entry` (since it is not just used for `domain_entry`)?

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -918,38 +918,80 @@ NCPInstanceBase::find_prefix_entry(const IPv6Prefix &prefix, const OnMeshPrefixE
...
 
-	iter = find_prefix_entry(prefix, entry);
+	if ((flags & OnMeshPrefixEntry::kFlagExtDP) == OnMeshPrefixEntry::kFlagExtDP)
+	{
+		iter = find_domain_prefix_entry(prefix, entry);
 
-	if (iter == mOnMeshPrefixes.end()) {
+		if (iter != mOnMeshPrefixes.end())
+		{
+			syslog(LOG_INFO, "OnMeshPrefixes: Removing old domain prefix %s", entry.get_description(prefix).c_str());
```
abtink, line 955: I think here in the log it should be from `iter` and not the `entry`?

src/wpantund/NCPInstanceBase-Addresses.cpp
```diff
@@ -918,38 +918,80 @@ NCPInstanceBase::find_prefix_entry(const IPv6Prefix &prefix, const OnMeshPrefixE
...
+	}
+
+	if (add_new_domain_entry)
+	{
 		mOnMeshPrefixes.insert(std::make_pair(prefix, entry));
 		syslog(LOG_INFO, "OnMeshPrefixes: Adding %s", entry.get_description(prefix).c_str());
 
-		if (origin != kOriginThreadNCP) {
+		if (origin != kOriginThreadNCP)
+		{
```
abtink, line 975: `wpantund` traditionally followed the style where `{` starts on the same line. 
We don't have or enforce any specific style guide for it though. 
I would suggest no need to change existing code. New code we can be more flexible.

XPAWAN:
@abtink all remarks has been applied and PR has been rebased. Kindly please review.

XPAWAN:
@abtink please do not merge this PR yet. I have some other additional improvement for this issue.

XPAWAN:
@abtink can You please review my changes. There is added checking if the on-mesh domain prefix is configured while adding a new one. To add a new on-mesh domain-prefix the previous one must be removed.

abtink (approved):
This is even better way. LGTM. Thanks. 👍

PR #492, opened 2021-04-22
NCP(USB CDC ACM) hard reset support

lmaciejonczyk:
This commit handles the situation when NCP(USB CDC ACM) performs hard reset after receiving SPINEL_CMD_RESET.

When NCP is connected by USB CDC ACM and we reset it, the NCP device is reattached to the host operating system. So far for NCP(USB CDC ACM) pseudo reset has been performing meaning only specific parts of firmware were reset. It is error-prone solution for bare metal device and almost unreachable for devices working under RTOS. But what's most important it does not guarantee the clean init state of the NCP device. It's far easier and better to hard reset NCP and properly handle the USB connection on the host side.

It uses udev library to check if device enumeration has ended.

It works based on symlink to the NCP device created by udev in /dev/serial/by-id/ directory.

### DIFF
--- a/src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
+++ b/src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
@@ -538,6 +538,16 @@ SpinelNCPInstance::driver_to_ncp_pump()
 		mOutboundBufferSent += pt->byte_count;
 #endif
 
+#if HAVE_LIBUDEV
+		uint8_t header = 0;
+		unsigned int command = 0;
+		(void)spinel_datatype_unpack(mOutboundBuffer, mOutboundBufferLen, "CiD", &header,
+					     &command, NULL, 0);
+		if (command == SPINEL_CMD_RESET) {
+			hard_reset_ncp();
+		}
+#endif
+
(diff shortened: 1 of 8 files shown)


src/ncp-spinel/SpinelNCPInstance-DataPump.cpp
```diff
@@ -538,6 +538,16 @@ SpinelNCPInstance::driver_to_ncp_pump()
 		mOutboundBufferSent += pt->byte_count;
 #endif
 
+#ifdef USB_CDC_ACM_RESET
```
abtink, line 541: Can we consider using`#if` check instead of `#ifdef`? 

I think in general we use (and prefer) `#if` style check in `wpantund` (allows all build configs to be defined in a common header file and the ones not needed to be set to zero).

abtink (approved):
Looks good overall. Thanks. 
One suggestion below

lmaciejonczyk:
@abtink I've addressed your comment, I hope this is what you expected. Is there anything I can do more here?

abtink:
Thanks @lmaciejonczyk . LGTM. I trigger a CI run on branch. Will merge after it passes.

Issue #494, opened 2021-05-14
Unable to forward NCP logs


superwhd:
**Describe the bug**
I tried to set up an NCP on nrf52840 and use wpantund to forward logs to syslog. However, I don't see NCP logs in `/var/log/syslog`.

**To Reproduce**
I was following the steps on https://github.com/openthread/ot-docs/blob/main/site/en/guides/build/logs.md#view-logs-for-an-ncp with some differences: 
- I was using `cmake` instead of `make`.
- The option `OPENTHREAD_CONFIG_LOG_OUTPUT_NCP_SPINEL` is not available any more so I used `OPENTHREAD_CONFIG_LOG_OUTPUT_APP`.

So here are my steps:
1. In repository ot-nrf528xx, I changed the log option at `src/nrf52840/openthread-core-nrf52840-config.h`
```
 #ifndef OPENTHREAD_CONFIG_LOG_OUTPUT
-#define OPENTHREAD_CONFIG_LOG_OUTPUT OPENTHREAD_CONFIG_LOG_OUTPUT_PLATFORM_DEFINED
+#define OPENTHREAD_CONFIG_LOG_OUTPUT OPENTHREAD_CONFIG_LOG_OUTPUT_APP
 #endif
```
2. I built NCP and flashed it to an nrf52840 board:
```
./script/build nrf52840 UART_trans -DOT_JOINER=ON -DOT_FULL_LOGS=ON
```
3. In `/etc/wpantund.conf`, I set `Daemon:SyslogMask "all"`.
4. I started wpantund as said on https://github.com/openthread/wpantund/blob/master/INSTALL.md#2-start-wpantund:
```
sudo /usr/local/sbin/wpantund -o NCPSocketName /dev/ttyACM0 
```
5. I ran `tail -F /var/log/syslog`.
6.  I used `wpanctl` to form a network. 
```
sudo /usr/local/bin/wpanctl form abc
```
7. In the terminal printing `/var/log/syslog`, I could see NCP messages forwarded by wpantund but I don't see NCP logs.

(I've manually removed some irrelevant rows/columns from the log file for privacy reasons.)

[wpantundlog.txt](https://github.com/openthread/wpantund/files/6477649/wpantundlog.txt)

[syslog.txt](https://github.com/openthread/wpantund/files/6477633/syslog.txt)

abtink:
[...]

PR #495, opened 2021-05-17
[doc] update INSTALL.md to use the latest master instead of full/latest-release

superwhd:
As mentioned in https://github.com/openthread/wpantund/issues/494, the INSTALL.md is outdated. For now we need to build and install `wpantund` using the latest master branch instead of `full/latest-release`, otherwise the NCP logs won't be forwarded properly.


abtink (approved):
Thanks. 👍

PR #518, opened 2024-01-04
[toranj] update the CI ubuntu environment

Irving-cl:
When CI ran for PR: https://github.com/openthread/wpantund/actions/runs/7385566082

It failed an reported a warning:
Screenshot 2024-01-04 at 13 42 54

I think this may be the root cause. This PR updates the ubuntu environment of CI.

### DIFF
--- a/.github/workflows/toranj.yml
+++ b/.github/workflows/toranj.yml
@@ -41,7 +41,7 @@ jobs:
       if: "github.ref != 'refs/heads/master'"
 
   toranj-ncp:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     env:
       TORANJ_RADIO : 15.4
     steps:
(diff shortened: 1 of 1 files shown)


Irving-cl:
Work in progress.

Irving-cl:
I've found the cause of failure when running toranj-ncp.

We should add env `top_builddir` so that in `wpan.py` it can finds the right path of ot-ncp-ftd. Otherwise the path is `../../..` and wpantund locates in /usr/local/sbin. The path is incorrect in this case.

Irving-cl:
Can get the CI work now.

Somehow the path prefix (`../../..`) of NCP used in `wpan.py` cannot work. I used a workaround to set environment variable `top_builddir` as the path obtained by running `pwd -P` in openthread directory.

abtink (approved):
Thanks. LGTM. 👍

jwhui (approved):

